PR #2, opened 2017-05-31
add deflate, add other function calls from the api, fix tag naming

mfurlotti:
I added some api calls relevant to create compressed data sets. I also fixed the field name so it is extracted from the relevant part of the tag, rather than getting set to the full tag string.

### DIFF
--- a/h5d.go
+++ b/h5d.go
@@ -177,11 +177,71 @@ func (s *Dataset) OpenAttribute(name string) (*Attribute, error) {
 	return openAttribute(s.id, name)
 }
 
+//H5_DLL hid_t H5Dget_space(hid_t dset_id);
+func (s *Dataset) H5Dget_space() (*Dataspace, error) {
+	dspace_id := (C.H5Dget_space(s.id))
+	ds := newDataspace(dspace_id)
+	if ds.id < 0 {
+		return (ds), fmt.Errorf("couldn't open dataspace from Dataset %q", s.Name())
+	}
+	return (ds), nil
+}
+
+//H5_DLL herr_t H5Dget_space_status(hid_t dset_id, H5D_space_status_t *allocation);
+
 // Datatype returns the HDF5 Datatype of the Dataset
+//H5_DLL hid_t H5Dget_type(hid_t dset_id);
 func (s *Dataset) Datatype() (*Datatype, error) {
 	dtype_id := C.H5Dget_type(s.id)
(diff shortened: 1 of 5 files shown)


h5d.go
```diff
@@ -177,11 +177,71 @@ func (s *Dataset) OpenAttribute(name string) (*Attribute, error) {
 	return openAttribute(s.id, name)
 }
 
+//H5_DLL hid_t H5Dget_space(hid_t dset_id);
+func (s *Dataset) H5Dget_space() (*Dataspace, error) {
+	dspace_id := (C.H5Dget_space(s.id))
```
kortschak: It looks like there are residual conversion parens. Can you clean these up please.

h5d.go
```diff
@@ -177,11 +177,71 @@ func (s *Dataset) OpenAttribute(name string) (*Attribute, error) {
 	return openAttribute(s.id, name)
 }
 
+//H5_DLL hid_t H5Dget_space(hid_t dset_id);
```
kortschak: What are these comments for?

h5t_shim.go
```diff
@@ -261,6 +261,7 @@ var (
 
 //
 var h5t_VARIABLE int64 = C.H5T_VARIABLE
+var H5S_UNLIMITED int64 = C.H5S_UNLIMITED
```
kortschak: Lead with lower case please.

h5d.go
mfurlotti: It is the prototype of the function being mapped, from the c header.

h5t_shim.go
mfurlotti: It needs to be exported. For instance, is passed to hdf5.CreateSimpleDataspace() to create an unlimited dataspace.

h5d.go
kortschak: Sure, that's what it is, but why is it here? There should be user-facing documentation (possibly, but by no means guaranteed including such a line) - think about how godoc renders this.

h5d.go
```diff
@@ -177,11 +177,69 @@ func (s *Dataset) OpenAttribute(name string) (*Attribute, error) {
...
+func (s *Dataset) H5Dget_space() (*Dataspace, error) {
```
sbinet: no need to namespace with `H5D`: it's already a method of `Dataset`.

PR #4, opened 2017-08-08
all: migrate to gonum.org/v1/hdf5

sbinet:
This CL migrates github.com/gonum/hdf5 to gonum.org/v1/hdf5.
It also makes sure people import it from that new location.

Fixes gonum/hdf5#3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,8 @@ matrix:
  allow_failures:
    - go: master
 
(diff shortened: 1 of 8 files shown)


sbinet:
@kortschak PTAL

kortschak:
The failure is only for tip `h5t_shim.go:263: constant 18446744073709551615 overflows int64`.

PR #5, opened 2017-08-09
travis: add go1.8, allow master to fail

sbinet:


.travis.yml
```diff
@@ -3,7 +3,12 @@ go:
   - 1.5
```
kortschak: While you are here, can you make these all `.x` (`- 1.5.x`) so that the highest patch version is tested.

sbinet: done

Issue #8, opened 2017-08-10
all: consider renaming Datatype into DataType


sbinet:
this was mentionned in gonum/hdf5#1:

we should probably use a better, Go-conforming, capitalization scheme for `Datatype` and rename it into `DataType`.

kortschak:
SGTM

PR #9, opened 2017-08-10
all: rename Datatype into DataType

sbinet:
Fixes gonum/hdf5#8.


kortschak:
Sorry, I'm having second thoughts about this change. I'm thinking about how much we can lend from type/kind naming in the reflect package; under this we would have `Datatype`=>`Type` and `Class`=>`Kind`. The single bit of grit that is getting in the way is that we have a type-like type associated with the `Identifier` type, currently called `IType`. I think maybe that should become `IdentifierType` (much as the long name is unpleasant). While that's happening, the `Id()` method should become `ID()` There are also a few cases where `N*` is used where `Num*` would be preferable.

PR #10, opened 2017-08-10
hdf5: fix build for >go1.8

kortschak:
@sbinet Please take a look.

### DIFF
--- a/h5t.go
+++ b/h5t.go
@@ -188,7 +188,7 @@ func (t *Datatype) Size() uint {
 }
 
 // SetSize sets the total size of a Datatype.
-func (t *Datatype) SetSize(sz uint) error {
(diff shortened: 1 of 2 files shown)


h5t.go
```diff
@@ -188,7 +188,7 @@ func (t *Datatype) Size() uint {
 }
 
 // SetSize sets the total size of a Datatype.
-func (t *Datatype) SetSize(sz uint) error {
+func (t *Datatype) SetSize(sz int) error {
```
kortschak: Note that this is incorrect for 32 bit systems (but so was the previous code).

kortschak: Actually, on a 32 bit arch, `int` and `size_t` are probably still the same size (though not guaranteed). I think we may want to have some init checks for things like this at some stage.

PR #14, opened 2018-02-22
Add Support for PacketTable Append

TacoVox:
As you guys stated on your front page, the PT interface was/is broken.
With a colleague, I went ahead and improved to things:
-  Removed `H5Pcreate_fl` and added the `H5Pcreate` as this is marked as deprecated by the official HDF5 documentation. As a matter of fact, I faced various issues, when trying to use variable lenght strings with `H5Pcreate_fl`.
- Fixed the Append function. There were countless issues with pointers and with the correct value representation in C that were fixed. I think the code will speak for itself.

Please consider this pull request.

### DIFF
--- a/h5pt.go
+++ b/h5pt.go
@@ -6,6 +6,7 @@ package hdf5
 
 // #include "hdf5.h"
 // #include "hdf5_hl.h"
+// #include <stdint.h>
 // #include <stdlib.h>
 // #include <string.h>
 import "C"
@@ -81,40 +82,170 @@ func (t *Table) ReadPackets(start, nrecords int, data interface{}) error {
 	return h5err(err)
 }
 
-// Append appends packets to the end of a packet table.
-func (t *Table) Append(data interface{}) error {
-	rv := reflect.Indirect(reflect.ValueOf(data))
-	rt := rv.Type()
-	c_nrecords := C.size_t(0)
-	c_data := unsafe.Pointer(nil)
+func extractStructValues(rv reflect.Value, rt reflect.Type) (ptr unsafe.Pointer, err error) {
+	// Initially the memory will be allocated to 64 bytes. If more is required
+	// later, this will be increased.
+	memSize := C.size_t(64)
+	ptr = C.malloc(memSize)
 
-	switch rt.Kind() {
+	for i := 0; i < rv.NumField(); i++ {
+		var dataPtr unsafe.Pointer
+		var dataSize C.size_t
+		f := rv.Field(i)
+		ft := f.Type().Kind()
+		offset := C.size_t(rt.Field(i).Offset)
 
-	case reflect.Array:
-		c_nrecords = C.size_t(rv.Len())
-		c_data = unsafe.Pointer(rv.UnsafeAddr())
+		switch ft {
+		case reflect.Int8:
+			val := C.int8_t(int8(f.Int()))
+			dataPtr = unsafe.Pointer(&val)
+			dataSize = C.size_t(unsafe.Sizeof(val))
+		case reflect.Uint8:
+			val := C.uint8_t(uint8(f.Uint()))
+			dataPtr = unsafe.Pointer(&val)
+			dataSize = C.size_t(unsafe.Sizeof(val))
+		case reflect.Int16:
+			val := C.uint8_t(int16(f.Int()))
+			dataPtr = unsafe.Pointer(&val)
+			dataSize = C.size_t(unsafe.Sizeof(val))
+		case reflect.Uint16:
+			val := C.uint8_t(uint16(f.Uint()))
+			dataPtr = unsafe.Pointer(&val)
+			dataSize = C.size_t(unsafe.Sizeof(val))
+		case reflect.Int32:
+			val := C.uint8_t(int32(f.Int()))
+			dataPtr = unsafe.Pointer(&val)
+			dataSize = C.size_t(unsafe.Sizeof(val))
+		case reflect.Uint32:
+			val := C.uint8_t(uint32(f.Uint()))
+			dataPtr = unsafe.Pointer(&val)
+			dataSize = C.size_t(unsafe.Sizeof(val))
+		case reflect.Int64:
+			val := C.int64_t(int64(f.Int()))
+			dataPtr = unsafe.Pointer(&val)
+			dataSize = C.size_t(unsafe.Sizeof(val))
+		case reflect.Uint64:
+			val := C.uint64_t(uint64(f.Uint()))
+			dataPtr = unsafe.Pointer(&val)
+			dataSize = C.size_t(unsafe.Sizeof(val))
+		case reflect.Float32:
+			val := C.float(float32(f.Float()))
+			dataPtr = unsafe.Pointer(&val)
+			dataSize = C.size_t(unsafe.Sizeof(val))
+		case reflect.Float64:
+			val := C.double(float64(f.Float()))
+			dataPtr = unsafe.Pointer(&val)
+			dataSize = C.size_t(unsafe.Sizeof(val))
+		case reflect.Bool:
+			val := C.uchar(0)
+			if f.Bool() {
+				val = 1
+			}
+			dataPtr = unsafe.Pointer(&val)
+			dataSize = C.size_t(unsafe.Sizeof(val))
+		default:
+			err = fmt.Errorf("hdf5: Could not append struct member %s "+
+				"to PacketTable", ft)
+			return nil, err
+		}
 
-	case reflect.Slice:
-		c_nrecords = C.size_t(rv.Len())
-		slice := (*reflect.SliceHeader)(unsafe.Pointer(rv.UnsafeAddr()))
-		c_data = unsafe.Pointer(slice.Data)
-
-	case reflect.String:
-		c_nrecords = C.size_t(rv.Len())
-		str := (*reflect.StringHeader)(unsafe.Pointer(rv.UnsafeAddr()))
-		c_data = unsafe.Pointer(str.Data)
-
-	case reflect.Ptr:
-		c_nrecords = C.size_t(1)
-		c_data = unsafe.Pointer(rv.Elem().UnsafeAddr())
-
-	default:
-		c_nrecords = C.size_t(1)
-		c_data = unsafe.Pointer(rv.UnsafeAddr())
+		// If the earlier allocated memory is not enough, we will increase it
+		// by another 64 bytes.
+		if memSize < (dataSize + offset) {
+			memSize += 64
+			ptr = C.realloc(ptr, memSize)
+			C.memset(unsafe.Pointer(uintptr(ptr)+uintptr(offset)), 0, 64)
+		}
+		C.memcpy(unsafe.Pointer(uintptr(ptr)+uintptr(offset)), dataPtr, dataSize)
 	}
 
-	err := C.H5PTappend(t.id, c_nrecords, c_data)
-	return h5err(err)
+	return ptr, nil
+}
+
+// Append appends packets to the end of a packet table.
+func (t *Table) Append(data interface{}) (err error) {
+	rv := reflect.Indirect(reflect.ValueOf(data))
+	rp := reflect.Indirect(reflect.ValueOf(&data))
+	rt := rv.Type()
+	cNrecords := C.size_t(1)
+	cData := unsafe.Pointer(nil)
+
+	switch rt.Kind() {
+	case reflect.Slice, reflect.Array:
+		for i := 0; i < rv.Len(); i++ {
+			if err = t.Append(rv.Index(i).Interface()); err != nil {
+				return err
+			}
+		}
+		return nil
+
+	case reflect.Struct:
+		if cData, err = extractStructValues(rv, rt); err != nil {
+			return err
+		}
+		defer C.free(cData)
+
+	case reflect.String:
+		stringData := C.CString(rv.String())
+		defer C.free(unsafe.Pointer(stringData))
+		cData = unsafe.Pointer(&stringData)
+
+	case reflect.Ptr:
+		ptrVal := rp.Elem()
+		cData = unsafe.Pointer(&ptrVal)
+
+	case reflect.Bool:
+		val := C.uchar(0)
+		if data.(bool) {
+			val = 1
+		}
+		cData = unsafe.Pointer(&val)
+
+	case reflect.Int8:
+		val := C.int8_t(data.(int8))
+		cData = unsafe.Pointer(&val)
+
+	case reflect.Uint8:
+		val := C.uint8_t(data.(uint8))
+		cData = unsafe.Pointer(&val)
+
+	case reflect.Uint16:
+		val := C.uint16_t(data.(uint16))
+		cData = unsafe.Pointer(&val)
+
+	case reflect.Int16:
+		val := C.int16_t(data.(int16))
+		cData = unsafe.Pointer(&val)
+
+	case reflect.Int32:
+		val := C.int32_t(data.(int32))
+		cData = unsafe.Pointer(&val)
+
+	case reflect.Uint32:
+		val := C.uint32_t(data.(uint32))
+		cData = unsafe.Pointer(&val)
+
+	case reflect.Int64:
+		val := C.int64_t(data.(int64))
+		cData = unsafe.Pointer(&val)
+
+	case reflect.Uint64:
+		val := C.uint64_t(data.(uint64))
+		cData = unsafe.Pointer(&val)
+
+	case reflect.Float32:
+		val := C.float(data.(float32))
+		cData = unsafe.Pointer(&val)
+
+	case reflect.Float64:
+		val := C.double(data.(float64))
+		cData = unsafe.Pointer(&val)
+
+	default:
+		return fmt.Errorf("hdf5: PT Append does not support datatype (%s).", rt.Kind())
+	}
+
+	return h5err(C.H5PTappend(t.id, cNrecords, cData))
 }
 
 // Next reads packets from a packet table starting at the current index into the value pointed at by data.
--- a/h5pt_test.go
+++ b/h5pt_test.go
@@ -11,33 +11,104 @@ import (
 import "testing"
 
 const (
-	fname    string = "ex_table_01.h5"
-	tname    string = "table"
-	nrecords int    = 8
+	fname     string = "ex_table_01.h5"
+	tname     string = "table"
+	chunkSize int    = 10
+	compress  int    = 0
 )
 
 type particle struct {
-	// name        string  `hdf5:"Name"`      // FIXME(sbinet)
+	//name        string  `hdf5:"Name"`		 // FIXME(TacoVox)
+	vehicle_no  uint8   `hdf5:"Vehicle Number"`
+	sattelites  int8    `hdf5:"Sattelites"`
+	cars_no     int16   `hdf5:"Number of Cars"`
+	trucks_no   int16   `hdf5:"Number of Trucks"`
+	min_speed   uint32  `hdf5:"Minimum Speed"`
 	lati        int32   `hdf5:"Latitude"`
+	max_speed   uint64  `hdf5:"Maximum Speed"`
 	longi       int64   `hdf5:"Longitude"`
 	pressure    float32 `hdf5:"Pressure"`
 	temperature float64 `hdf5:"Temperature"`
+	accurate    bool    `hdf5:"Accurate"`
 	// isthep      []int                     // FIXME(sbinet)
 	// jmohep [2][2]int64                    // FIXME(sbinet)
 }
 
-func TestTable(t *testing.T) {
+func testTable(t *testing.T, dType interface{}, data interface{}, nrecords int) {
+	var table *Table
 
+	typeString := reflect.TypeOf(data).String()
+
+	// create a new file using default properties
+	f, err := CreateFile(fname, F_ACC_TRUNC)
+	if err != nil {
+		t.Fatalf("CreateFile failed for %s: %s", typeString, err)
+	}
+	defer os.Remove(fname)
+	defer f.Close()
+
+	table, err = f.CreateTableFrom(tname, dType, chunkSize, compress)
+	if err != nil {
+		t.Fatalf("CreateTableFrom struct failed for %s: %s", typeString, err)
+	}
+	defer table.Close()
+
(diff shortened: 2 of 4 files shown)


TacoVox:
I think the build fails, beacuse your system uses an outdated version of HDF5...

sbinet:
@TacoVox Thanks for having a stab at this!

h5pt.go
```diff
@@ -84,36 +90,82 @@ func (t *Table) ReadPackets(start, nrecords int, data interface{}) error {
...
+	case reflect.Int8:
+		val := C.char(data.(int8))
+		c_data = unsafe.Pointer(&val)
+
+	case reflect.Uint8:
+		val := C.uchar(data.(int8))
+		c_data = unsafe.Pointer(&val)
+
+	case reflect.Int32:
+		val := C.int(data.(int32))
```
sbinet: here, I'd rather use `C.int32_t`. what do you think?

h5pt.go
```diff
@@ -84,36 +90,82 @@ func (t *Table) ReadPackets(start, nrecords int, data interface{}) error {
...
+		c_data = unsafe.Pointer(&val)
+
+	case reflect.Uint32:
+		val := C.uint(data.(uint32))
```
sbinet: `C.uint32_t` ?

h5pt.go
```diff
@@ -84,36 +90,82 @@ func (t *Table) ReadPackets(start, nrecords int, data interface{}) error {
...
+		c_data = unsafe.Pointer(&val)
+
+	case reflect.Int64:
+		val := C.long(data.(int64))
```
sbinet: `C.int64_t` ?

h5pt.go
```diff
@@ -84,36 +90,82 @@ func (t *Table) ReadPackets(start, nrecords int, data interface{}) error {
...
+		c_data = unsafe.Pointer(&val)
+
+	case reflect.Uint64:
+		val := C.ulong(data.(uint64))
```
sbinet: `C.uint64_t` ?

h5pt.go
```diff
@@ -84,36 +90,82 @@ func (t *Table) ReadPackets(start, nrecords int, data interface{}) error {
...
+	case reflect.Float32:
+		val := C.float(data.(float32))
+		c_data = unsafe.Pointer(&val)
+
+	case reflect.Float64:
+		val := C.double(data.(float64))
+		c_data = unsafe.Pointer(&val)
+
+	case reflect.Complex64:
+		val := C.complexfloat(data.(complex64))
```
sbinet: for this to compile, we'd need to add `#include <complex.h>` in the `import "C"` at the top of the file.

h5pt.go
```diff
@@ -84,36 +90,82 @@ func (t *Table) ReadPackets(start, nrecords int, data interface{}) error {
...
+		c_data = unsafe.Pointer(&val)
+
+	case reflect.Complex128:
+		val := C.complexdouble(data.(complex128))
```
sbinet: ditto.

h5pt.go
```diff
@@ -84,36 +90,82 @@ func (t *Table) ReadPackets(start, nrecords int, data interface{}) error {
...
+		c_data = unsafe.Pointer(&val)
 
 	default:
-		c_nrecords = C.size_t(1)
-		c_data = unsafe.Pointer(rv.UnsafeAddr())
+		return fmt.Errorf("PT Append does not support datatype (%s).", rt.Kind())
```
sbinet: please prepend `"hdf5: ..."` to the error message.

h5pt.go
```diff
@@ -20,10 +20,11 @@ import (
 // Table is an hdf5 packet-table.
 type Table struct {
 	Identifier
+	strings []unsafe.Pointer
 }
 
 func newPacketTable(id C.hid_t) *Table {
-	t := &Table{Identifier{id}}
+	t := &Table{Identifier{id}, make([]unsafe.Pointer, 10)}
```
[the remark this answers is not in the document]
TacoVox: Arbitrary number. We can obviously set this to 1, or whatever you think is better.

h5pt.go
```diff
@@ -20,10 +20,11 @@ import (
 // Table is an hdf5 packet-table.
 type Table struct {
 	Identifier
+	strings []unsafe.Pointer
```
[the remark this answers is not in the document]
TacoVox: Sure :)

h5pt.go
TacoVox: Better.

h5pt.go
TacoVox: +1

h5pt.go
TacoVox: +1

h5pt.go
TacoVox: +1

h5pt.go
TacoVox: Oh. I forgot that. :-1:

h5pt.go
TacoVox: Sure.

TacoVox:
Improved based on your comments. I think it is best, if you would update your hdf5 version to 1.10.1, as it comes with quite some improvements.

sbinet:
unfortunately, `trusty` has only `1.8.11`: https://packages.ubuntu.com/trusty/libhdf5-7

perhaps you could split that PR in 2:
- first part w/o the new `H5PTcreate`
- a second one with the new API (so we can setup travis to compile the new hdf5 version)

what do you think?

TacoVox:
I know that this is the case for Ubuntu based systems. Would it be hard to compile it yourself? 
But yes sure, I could split this pull request into two ;)

sbinet:
in my experience, it takes a few iterations to wrangle with travis+local-compilations :)
hence my proposal to split the PR in 2 (so the easy part can be merged sooner)

TacoVox:
One last thing, we need to revert back to unsafe.Pointer, as there are different issues arising:
- H5PTappend wants a void pointer --> unsafe.Pointer
- C.free() as well
I could cast, but I think unsafe.Pointer is a nicer solution.

TacoVox:
I see that the test fails in regard to structs. I would suggest that Append should not support structs. I do not know if you agree. I just think that it would be quite nasty. I would adapt the code to iterate through the struct members, reflecting again on their type, make a recursive call. I think that is too nasty.
What you do you think?
If you agree, I would adapt the tests for testing for specific datatypes, which I think is more valid.

sbinet:
I think I'd prefer the better "typed" `[]*C.char` (and thus, the `unsafe.Pointer` cast).
(but I'll let @kortschak have a say on this :P)

I do agree that, even if I wouldn't use "nasty", supporting structs with members that are (or contain) pointers to other Go-allocated values is more challenging.
Let's table (pun intended) this for later.

kortschak:
FWIW I think a `[]*C.char` is preferable.

TacoVox:
Alright, but I would like to understand your reasoning behind that, as it will make the code a bit more complex.

sbinet:
`cgo` isn't C not is it Go.
So you get the worst of both languages :)

Having slightly more typed variable may help to prevent future mistakes.
Re-converting to `unsafe.Pointer` to appease the C API is (IMHO) a minor inconvenience.

(Again, thanks for working on this)

TacoVox:
Okidoki. I buy that 👍 ;)
I will implement this and update the tests on Monday and push then.
Maybe I will find some time later to update in regard to structs. But that would then be a seperate PR.
And no worries - it's fun :)

sbinet:
SGTM

TacoVox:
So I updated the stuff.
I ditched the complex stuff, as there is - IMHO - not a proper corresponding datatype for HDF5.
Nevertheless, I added some stuff for bool in h5t.go.
The test will still fail. But I assume that this comes from line 70 in the test file. I think there is an issue with CreateTableFrom.

TacoVox:
Which is kinda obvious: you will need to set one type for each packet table. It cannot have several. I will update the test code accordingly.

TacoVox:
Actually, I am curious: how did you guys vision the Append of structs in a PacketTable.
I certainly haven't used it as such. I always saw a PT as a "one datatype only" thing, and so far I see nothing that contradicts that (e.g. https://support.hdfgroup.org/HDF5/doc/HL/H5PT_Intro.html).
Therefore, I do not see any way in allowing structs to be added to this.
I will wait with further pushes until we are on the same page, I guess :)

sbinet:
the way I'd envisionned it (and it was kinda working pre-Go1.6) was to create a new data type associated to a Go struct, using `reflect` to introspect that Go type to create the equivalent hdf5 data type.
that's done in:
https://github.com/TacoVox/hdf5/blob/af0067f662b834eef0f15506dc5c6c661a179b7c/h5pt.go#L242

TacoVox:
So in case you have a struct like
```
type mys struct {
    field1 string
    field2 int64
    field3 float64
}
```
You want to create three PacketTables  each with the corresponding type? As PacketTable must have one specific type, right?

sbinet:
no, one would create an hdf5 `CompoundType` mirroring that `mys` struct.
The PacketTable would thus have 1 specific, homogenous type: `{string, int64, float64}`.
see:
http://santoshdarekar-hdf5.blogspot.fr/2011/02/compound-data-in-packet-table.html

except that starting with Go-1.6, we aren't allowed anymore to give to the C side a pointer to a Go value that itself contains pointer(s) to other Go values.
A Go value of type `mys` has a (hidden) pointer: a `string` value is `(len, ptr-to-actual-data)`, so we can't just give it to the C-side anymore...

kortschak:
ISTM that the solution to the both is making a reasonably generic deep copy function that will copy a Go data structure into a `C.malloc`ed memory are (and back for roundtripping) to hand over to hdf5. This is unfortunate, but apparently the only reasonable approach.

TacoVox:
Exactly. Sporadically working with that (not so nice) approach since yesterday. Will keep you posted.

sbinet:
FYI, I have written something along these lines there:
https://github.com/go-hep/croot/tree/master/cmem

TacoVox:
As you can see, I pushed a version that is not working. I hope to get some help from you guys. I do not see where it fails. No matter if I set `c_data` to `unsafe.Pointer(ptr)` `unsafe.Pointer(&ptr)` or the current, I get either a failure while writing, or an empty table.

h5pt.go
```diff
@@ -84,36 +96,149 @@ func (t *Table) ReadPackets(start, nrecords int, data interface{}) error {
...
+				data_size = C.size_t(1)
+			default:
+				return fmt.Errorf("hdf5: Could not append struct member %s "+
+					"to PacketTable", ft)
+			}
+
+			ptr = C.realloc(ptr, ptr_offset+data_size)
+			C.memset(unsafe.Pointer(uintptr(ptr)+uintptr(ptr_offset)), 0, data_size)
+			C.memcpy(unsafe.Pointer(uintptr(ptr)+uintptr(ptr_offset)), data_ptr, data_size)
+			ptr_offset += data_size
```
sbinet: this doesn't take into account C memory layout+alignment.
ie: each platform has its own memory alignment rules (mainly depending on 32b/64b)

when we pass a `void*` to hdf5, pointing at a C-struct, hdf5 expects that C-struct to honour the C alignment rules.

```c
struct S {
    uint32_t f1; // -> offset=0
    uint16_t f2; // -> offset=4
    uint8_t  f3; // -> offset=6
    uint16_t f4; // -> offset=8
};
```

this will have `sizeof(S) == 12` (on my 64b machine) and not 9.

TacoVox:
So this version now works for go 1.7+. Do you guys need support for versions below that?
We changed the test code so that it hopefully covers every case. Having said that, strings within a struct are excluded. We are not yet sure if that is even properly possible as the official HDF5 documentation (https://support.hdfgroup.org/HDF5/Tutor/compound.html) states that it has 
> It has a fixed total size, in bytes

Furthermore, I am looking into improving (*Table) readPackets() as it seems not to work as expected.

kortschak:
Please feel free to drop <go1.7 (and add go1.10); we only support the most recent three releases of Go.

Thank you for putting all this work in.

TacoVox:
As commit states, I cleaned up the code. Fixed a memory leak and changed the test a bit so that it will still test for original the particle case.
This is marked as FIXME, as someone has to improve the ReadTable or just the part of the tests that test for it. I would prefer that this will be handled in another PR.
What you say @sbinet?

h5pt.go
```diff
@@ -81,40 +93,172 @@ func (t *Table) ReadPackets(start, nrecords int, data interface{}) error {
...
+		cData = unsafe.Pointer(&val)
+
+	case reflect.Float64:
+		val := C.double(data.(float64))
+		cData = unsafe.Pointer(&val)
 
 	default:
-		c_nrecords = C.size_t(1)
-		c_data = unsafe.Pointer(rv.UnsafeAddr())
+		err = fmt.Errorf("hdf5: PT Append does not support datatype (%s).", rt.Kind())
```
sbinet: if we take this branch of the `switch`, it seems to me like we drop this error on the floor as it's overriden later on, at line 259.

h5pt_test.go
```diff
@@ -5,109 +5,261 @@
...
 	longi       int64   `hdf5:"Longitude"`
 	pressure    float32 `hdf5:"Pressure"`
 	temperature float64 `hdf5:"Temperature"`
+	accurate    bool    `hdf5:"Accurate"`
 	// isthep      []int                     // FIXME(sbinet)
 	// jmohep [2][2]int64                    // FIXME(sbinet)
 }
 
-func TestTable(t *testing.T) {
+func testTable(dType interface{}, data interface{}, nrecords int) (err error) {
```
sbinet: `testTable` should probably take a `t *testing.T` parameter instead of return an error.

this would make for a better debugging experience: calling `t.Errorf` or `t.Fatalf` instead of return an error and thus getting a full stack trace.

sbinet:
@TacoVox yes, I'd be happy with submitting something, even if with a few FIXMEs left for later.

h5pt_test.go
TacoVox: Hmm yeah. Maybe you are right. Will adapt it accordingly.

h5pt.go
TacoVox: Yes. You are right. Will need to update the returns.

h5pt.go
```diff
@@ -20,10 +21,12 @@ import (
 // Table is an hdf5 packet-table.
 type Table struct {
 	Identifier
+	strings []*C.char
+	structs []unsafe.Pointer
 }
 
 func newPacketTable(id C.hid_t) *Table {
-	t := &Table{Identifier{id}}
+	t := &Table{Identifier{id}, make([]*C.char, 1), make([]unsafe.Pointer, 1)}
```
sbinet: I don't understand why we create the 2 slices with a length of 1.
it doesn't seem like we do `t.structs[0] = ... ` nor `t.strings[0] = ... `
it seems we always `append` to these slices.

maybe just leave them default-initialized?

h5pt.go
```diff
@@ -81,40 +93,170 @@ func (t *Table) ReadPackets(start, nrecords int, data interface{}) error {
...
+	ptr = C.malloc(memSize)
+
+	for i := 0; i < rv.NumField(); i++ {
+		var dataPtr unsafe.Pointer
+		var dataSize C.size_t
+		f := rv.Field(i)
+		ft := string(f.Type().Name())
+		offset := C.size_t(rt.Field(i).Offset)
+
+		switch ft {
```
sbinet: perhaps use `ft := f.Type().Kind()` and switch on `reflect.Int8`, `reflect.Uint8`, ... instead of switching on and comparing strings ?

h5pt.go
```diff
@@ -81,40 +93,170 @@ func (t *Table) ReadPackets(start, nrecords int, data interface{}) error {
...
 	rv := reflect.Indirect(reflect.ValueOf(data))
+	rp := reflect.Indirect(reflect.ValueOf(&data))
 	rt := rv.Type()
-	c_nrecords := C.size_t(0)
-	c_data := unsafe.Pointer(nil)
+	cNrecords := C.size_t(1)
+	cData := unsafe.Pointer(nil)
 
 	switch rt.Kind() {
+	case reflect.Slice, reflect.Array:
```
sbinet: thinking a bit more about this, I realize these new semantics break what (I think) the concept of a packet means for hdf5.

it seems to me it is a valid hdf5 thing to say you want a packet to be 1 slice or 1 array of N float64s.
doing it like we do it now, we unravel each packet of 1 array of N float64s into N packets of 1 float64.

I believe the handling of arrays and slices can be left at it was before.

h5pt_test.go
```diff
@@ -11,103 +11,192 @@ import (
...
+	table, err = f.CreateTableFrom(tname, dType, chunkSize, compress)
 	if err != nil {
-		t.Fatalf("CreateTableFrom failed: %s", err)
+		t.Fatalf("CreateTableFrom struct failed for %s: %s", typeString, err)
 	}
 	defer table.Close()
 
 	if !table.IsValid() {
-		t.Fatal("table is invalid")
+		t.Fatalf("PacketTable \"%s\" is invalid for %s.", tname, typeString)
```
sbinet: `\"%s\"` can be replaced with `%q`.

h5pt.go
```diff
@@ -41,6 +44,15 @@ func (t *Table) Close() error {
 	}
 	err := h5err(C.H5PTclose(t.id))
 	t.id = 0
+
+	for _, ptr := range t.strings {
```
sbinet: I haven't managed to get a clear spec/promise from the hdf5 docs, but it seems (experiment-programming) that `H5PTappend` doesn't hold onto the memory one passes to it (our `cData` variable in `Append`).

so, the lifetime of `t.strings` and `t.structs` could be restricted to just `Append`:
- (re)allocating them
- freeing them
could all be done inside `Append` and allow for more reasonnable memory usage during long-running hdf5.Table appending.

h5pt.go
TacoVox: I would've love to keep your code. But it does not pass the test:

> --- FAIL: TestPTStruct (0.00s)
> panic: reflect.Value.UnsafeAddr of unaddressable value [recovered]
>         panic: reflect.Value.UnsafeAddr of unaddressable value
> 
> goroutine 12 [running]:
> testing.tRunner.func1(0xc4200be780)
>         /snap/go/1473/src/testing/testing.go:742 +0x29d
> panic(0x52b920, 0x5797a0)
>         /snap/go/1473/src/runtime/panic.go:505 +0x229
> reflect.Value.UnsafeAddr(0x5281e0, 0xc42000c180, 0x97, 0x7)
>         /snap/go/1473/src/reflect/value.go:1772 +0xc4
> hdf5_write/vendor/gonum.org/v1/hdf5.(*Table).Append(0xc420026800, 0x5281e0, 0xc42000c180, 0x553740, 0xc42001a200)
>         /home/jonas/Repos/rapiddecode/write/src/hdf5_write/vendor/gonum.org/v1/hdf5/h5pt.go:199 +0x709
> hdf5_write/vendor/gonum.org/v1/hdf5.testTable(0xc4200be780, 0x553740, 0xc42001a200, 0x5281e0, 0xc42000c180, 0x7)
>         /home/jonas/Repos/rapiddecode/write/src/hdf5_write/vendor/gonum.org/v1/hdf5/h5pt_test.go:61 +0x3d9
> hdf5_write/vendor/gonum.org/v1/hdf5.TestPTStruct(0xc4200be780)
>         /home/jonas/Repos/rapiddecode/write/src/hdf5_write/vendor/gonum.org/v1/hdf5/h5pt_test.go:127 +0x200
> testing.tRunner(0xc4200be780, 0x5689d8)
>         /snap/go/1473/src/testing/testing.go:777 +0xd0
> created by testing.(*T).Run
>         /snap/go/1473/src/testing/testing.go:824 +0x2e0
> FAIL    hdf5_write/vendor/gonum.org/v1/hdf5     0.008s

h5pt.go
TacoVox: No problem :) Fixed

h5pt_test.go
TacoVox: Good point :+1:

h5pt.go
TacoVox: We actually had so many issues with that in the initial development. Therefore, we introduced those pointer slices. BUT: I tried now again - and it just worked fine. :+1:

h5pt.go
TacoVox: With the change you suggest below, this loses relevance :) But thanks for pointing out!

TacoVox:
Thanks for pointing out those things. I fixed all of them, except for one (see above).
Cheers!

h5pt.go
```diff
@@ -41,6 +42,7 @@ func (t *Table) Close() error {
 	}
 	err := h5err(C.H5PTclose(t.id))
 	t.id = 0
+
```
sbinet: just for the sake of reducing line noise in diffs, could this be removed ?

TacoVox:
Wonderful :+1:  :)
I think this should do it :)

TacoVox:
And just FYI - I found some more things that I think need some fixes :P I do not think that this will be the last PR... :octocat:

sbinet:
ok.

I am not completely satisfied with the commit messages :)

could they look like:

```
travis: drop Go-1.{5,6,7}.x, add Go-1.10.x
```

and:
```
hdf5: fix H5PTappend to work with CGo-1.6 rules

- In order to add support for the append of slices and arrays, a
  recursive call to the Append function itself was introduced.
- To support strings, the Go string is now converted to a CString. The
  pointer of that CString is then used to write to the HDF5 file.
- To support pointers from arbitrary values, the reflect value of the
  address to that pointer was added.
- All supported primitive Go types are now converted to the
  corresponding C type to ensure that the values are represented
  correctly.
- To support structs, a new function was introduced that takes the
  struct's data and puts it into memory with the correct offsets.
- The test function was updated accordingly. It tests now for all of the
  above test cases, needs some improvement in regard to structs. A TODO
  mark was set.
```

(so, the bodies of the commit messages are ok, but, following the Go team rules for commit messages, we like to have a nice one-liner for commit messages headers.
see: https://github.com/gonum/gonum/blob/master/CONTRIBUTING.md
)

thanks a lot!

TacoVox:
Fixed :+1:

sbinet:
thanks!

the hard part was the first PR. now it's just happy and smooth sailing... :)
keep 'em coming.

PR #15, opened 2018-04-16
hdf: Add support for pointers in NewDataTypeFromType()

TacoVox:
This small patch introduces support for pointers to datatypes in the
NewDataTypeFromType function. This enables support for different use
cases, e.g. usage with google.com/gopacket

Co-authored-by: Gustaf Johansson <gustaf.j@gmail.com>

### DIFF
--- a/h5t.go
+++ b/h5t.go
@@ -466,6 +466,9 @@ func NewDataTypeFromType(t reflect.Type) (*Datatype, error) {
 		}
 		dt = &cdt.Datatype
 
+	case reflect.Ptr:
+		return NewDataTypeFromType(t.Elem())
+
 	default:
 		// Should never happen.
 		panic(fmt.Errorf("unhandled kind (%v)", t.Kind()))
--- a/h5t_test.go
+++ b/h5t_test.go
@@ -26,10 +26,13 @@ func TestSimpleDatatypes(t *testing.T) {
 		float32(0),
 		float64(0),
 		string(""),
+		bool(true),
 	}
 
 	for test := range tests {
 		NewDatatypeFromValue(test)
+		// Test again for usage with ptrs
+		NewDatatypeFromValue(&test)
 	}
 }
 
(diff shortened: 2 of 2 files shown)


h5t.go
```diff
@@ -466,6 +466,9 @@ func NewDataTypeFromType(t reflect.Type) (*Datatype, error) {
 		}
 		dt = &cdt.Datatype
 
+	case reflect.Ptr:
+		return NewDataTypeFromType(t.Elem())
```
sbinet: if `t` is `*int`, this will return the HDF5 DataType for `int`.
shouldn't we instead return the HDF5 DataType that describes a pointer to `int` ?

gustafj: This type specification is in line with the current implementation of Append:
https://github.com/gonum/hdf5/blob/master/h5pt.go#L193-L195
```
	case reflect.Ptr:
		ptrVal := rp.Elem()
		cData = unsafe.Pointer(&ptrVal)
```
If you propose to use `H5T_STD_REF_OBJ` as a type of `*int` and then add the actual `int` to a separate file location the `Append` function (possibly among many others) would need to be refactored as well.
The change as is now, prevents a panic in `default` and works as intended.

TacoVox: I fixed the test cases. I think they are testing now the stuff that should be tested.

TacoVox:
Have you had time to check it out @sbinet? :ghost:  I think there will be another patch coming in shortly, as well... Support for arrays/slices and nested struct appends. But I want to have that one closed first :smile:

h5t_test.go
```diff
@@ -65,48 +68,62 @@ func TestStructDatatype(t *testing.T) {
 
 	// Test that the type can be constructed and that the number of
 	// members is as expected.
+	dtypes := make([]*Datatype, 0)
```
sbinet: we usually spell this out as:
```go
var dtypes []*Datatype
```

TacoVox: If it is really pressing, I can obviously fix that one.

kortschak: Please do.

TacoVox:
Alrighty. Amended the commit. Is now aligned with your usual spelling.
Do I also need to update the contributors thing for @gustafj or can that wait for the next PR?

sbinet:
> Alrighty. Amended the commit. Is now aligned with your usual spelling.

thanks :)

> Do I also need to update the contributors thing for @gustafj or can that wait for the next PR?

If I am not mistaken, @gustafj did commit something in the first commit, so, yeah, both of you will need to send a PR against [gonum/gonum](https://github.com/gonum/gonum)

TacoVox:
As I am already in there, it will just be him. Will fix shortly :) Thx!

TacoVox:
PR added. + I fixed the commits. :)

TacoVox:
Just a quick reminder @sbinet  :))
Thanks once again that you are so open for the PRs!

sbinet:
Actually, we are now waiting on @kortschak :)

If he doesn't but by Monday (when I come back from holidays) I'll merge this in.

sbinet:
No worries.

PR #17, opened 2018-05-03
(H5PT) Improved Append() Method

TacoVox:
To keep the fun alive :sunglasses: 

This patch features an improvement of the HDF5 PacketTable Append
method. This was necessary, because the old version did not support the
append of struct containing nested structs and/or slices/arrays.
To sum it up:
- Any data of any Compound Datatype can be written to a PacketTable now.
- This is achieved by allocating memory that contains the necessary
  data.
- The whole code is simpler than the doubled code that excisted before.
- The Append() method now supports proper appending of slices, which are
  not appended in a append-per-member manner, but rather as several
  objects to append that are consecutively in memory. This is more
  aligned with the offical way-of-doing.

What needs to be improved?
- The append of pointers still follows the old notion that we derefence
  the pointer and write the value directly. To support H5T_STD_REF_OBJ,
  further, significant changes have to be done to the whole code base.
- The read functionality has to be revised, as it seems to not work when
  strings are part of compound datatypes. Manual testing has shown that
  they are written properly, the test cases still would fail, because
  of the mentioned call to read.

I suggest that the above mentioned improvements are tackled in a separate patch.

### DIFF
--- a/h5pt.go
+++ b/h5pt.go
@@ -18,6 +18,26 @@ import (
 	"unsafe"
 )
 
+// appendData is a wrapper type for information necessary to create and
+// subsequently write an in memory object to a PacketTable using Append().
+type appendData struct {
+	ptr        unsafe.Pointer
+	memSize    C.size_t
+	offset     C.size_t
+	numRecords C.size_t
+	strs       []unsafe.Pointer
+}
+
+func (ad *appendData) free() {
+	C.free(ad.ptr)
+	ad.ptr = nil
+
+	for _, str := range ad.strs {
+		C.free(str)
+		str = nil
+	}
+}
+
 // Table is an hdf5 packet-table.
 type Table struct {
 	Identifier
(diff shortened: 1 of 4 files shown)


TacoVox:
Yes, take your time :) I guess you will have numerous comments :sunglasses:

h5pt.go
```diff
@@ -23,6 +23,14 @@ type Table struct {
 	Identifier
 }
 
+type appendData struct {
+	ptr       unsafe.Pointer
+	memSize   C.size_t
+	offset    C.size_t
+	noRecords C.size_t
```
sbinet: I would rename this `numRecords` or `nrecords`.
`noRecords` reads as it's "no (such) record".

h5pt.go
```diff
@@ -82,170 +90,154 @@ func (t *Table) ReadPackets(start, nrecords int, data interface{}) error {
...
+	defer C.free(ad.ptr)
+
+	if err = extractValues(&ad, data); err != nil {
+		return
+	}
+
+	// Make sure that we free all memory that was allocated to strings after
+	// they were written to the HDF5 file.
+	for _, str := range ad.strs {
+		defer C.free(str)
```
sbinet: instead of `len(ad.strs)` `defer`s, couldn't we create only one `defer` whose body does the looping + freeing?

h5pt.go
```diff
@@ -82,170 +90,154 @@ func (t *Table) ReadPackets(start, nrecords int, data interface{}) error {
...
+	ad.offset += dataSize
+
+	return
+}
+
+// Append appends packets to the end of a packet table.
+func (t *Table) Append(data interface{}) (err error) {
+	ad := appendData{
+		ptr: unsafe.Pointer(nil),
+		// Ensure to start with zero initialization on curcial fields
```
sbinet: none of these are needed. in Go, everything is initialized with a zero value.
so:
```go
var ad appendData
```

h5pt.go
sbinet: actually, we could refactor all these `defer`s into one method on `appendData`:
```go
func (ad *appendData) free() {
    C.free(ad.ptr)
    for _, str := range ad.strs {
        C.free(str)
    }
}
```

so we could then just write:
```go
var ad appendData
defer ad.free()
```

h5pt.go
TacoVox: Good point. I actually added this for debugging purposes (when something fishy was going on) and never removed it. Thx for spotting.

h5pt.go
TacoVox: Nice! Will do that.

TacoVox:
Hi @sbinet,
yes you are right. It is not the most beautiful way of handling this. But for me it seems the best way of handling it ATM. Obviously, revising the code is always an option :)

Nevertheless, we are using the code quite successfully here for some huge datasets (appending 5MB+ a time at a high rate). And so far `Append()` was stable. I hope this makes you a bit more comfortable to hear :)

TacoVox:
Alright! But definitely something to keep in mind! :sunglasses:

TacoVox:
Actually there is one thing @sbinet. It just came to my mind that I have no case for `uint` and `int`. Do you want me to add one that is then just a `uint64` and `int64`?
Any other changes that you want for this PR?

sbinet:
a Go `int` has the size of `int64` only on a 64b machine.
this would make the creation of an `HDF5` file non portable...

I would err on the side of just returning an error.
(with a `CMarshal` version we could actually create a `[]byte` with `int64` data w/o fiddling w/ alignment+offset)

TacoVox:
Okay. Then we can keep it for now?
I guess so, but I would need to do that another time and PR :smiley:

sbinet:
LGTM but could you reword a bit the commit message so it reads something like:

```
hdf5: improve Table.Append bla-bla

bla-bla foo bar
qux and baz
```

TacoVox:
Done, sir :)

sbinet:
thanks!
let's wait a bit for @kortschak to chime in :)

h5pt.go
```diff
@@ -82,170 +97,143 @@ func (t *Table) ReadPackets(start, nrecords int, data interface{}) error {
...
-		}
-		C.memcpy(unsafe.Pointer(uintptr(ptr)+uintptr(offset)), dataPtr, dataSize)
-	}
-
-	return ptr, nil
-}
-
-// Append appends packets to the end of a packet table.
-func (t *Table) Append(data interface{}) (err error) {
+// Extracts the values that have to be appended to a PT. Writes them to mem, which
```
kortschak:
```
// extractValues extracts the values to be appended to a packet table and adds
// them to the appendData structure.
//
// Struct values must only have exported fields, otherwise extractValues will panic.
```

Or alternatively elide those. I think the current behaviour is OK.

h5pt.go
```diff
@@ -18,6 +18,21 @@ import (
 	"unsafe"
 )
 
+type appendData struct {
```
kortschak: Can we have some documentation on this please.

h5pt_test.go
```diff
@@ -18,20 +18,20 @@ const (
...
-	max_speed   uint64  `hdf5:"Maximum Speed"`
-	longi       int64   `hdf5:"Longitude"`
-	pressure    float32 `hdf5:"Pressure"`
-	temperature float64 `hdf5:"Temperature"`
-	accurate    bool    `hdf5:"Accurate"`
-	// isthep      []int                     // FIXME(sbinet)
-	// jmohep [2][2]int64                    // FIXME(sbinet)
+	// Name        string  `hdf5:"Name"` // FIXME(TacoVox): ReadPackets seems to need an update
+	Vehicle_no  uint8   `hdf5:"Vehicle Number"`
+	Sattelites  int8    `hdf5:"Sattelites"`
```
kortschak: Satellites

h5pt.go
```diff
@@ -82,170 +97,143 @@ func (t *Table) ReadPackets(start, nrecords int, data interface{}) error {
...
-	}
-
-	return ptr, nil
-}
-
-// Append appends packets to the end of a packet table.
-func (t *Table) Append(data interface{}) (err error) {
+// Extracts the values that have to be appended to a PT. Writes them to mem, which
+// is then written to the HDF5 file.
+func extractValues(ad *appendData, data interface{}) (err error) {
```
kortschak: A named return is not necessary here.

h5pt.go
```diff
@@ -82,170 +97,143 @@ func (t *Table) ReadPackets(start, nrecords int, data interface{}) error {
...
 		for i := 0; i < rv.Len(); i++ {
-			if err = t.Append(rv.Index(i).Interface()); err != nil {
-				return err
+			if err = extractValues(ad, rv.Index(i).Interface()); err != nil {
+				return
 			}
 		}
-		return nil
+		ad.numRecords = C.size_t(rv.Len())
+		return
```
kortschak: Please don't use naked returns like this; make this and others like it, `return nil` if nil, otherwise `return err`.

h5pt.go
```diff
@@ -82,170 +97,143 @@ func (t *Table) ReadPackets(start, nrecords int, data interface{}) error {
...
+
+	ad.memSize = dataSize + ad.offset
+	ad.ptr = C.realloc(ad.ptr, ad.memSize)
+	C.memcpy(unsafe.Pointer(uintptr(ad.ptr)+uintptr(ad.offset)), dataPtr, dataSize)
+	ad.offset += dataSize
+
+	return
+}
+
+// Append appends packets to the end of a packet table.
```
kortschak: Add comment similar to the one above about visibility of fields and remove the named return.

h5pt.go
```diff
@@ -82,170 +97,143 @@ func (t *Table) ReadPackets(start, nrecords int, data interface{}) error {
...
+func (t *Table) Append(data interface{}) (err error) {
+	var ad appendData
+	defer ad.free()
+
+	if err = extractValues(&ad, data); err != nil {
```
kortschak:
```
err := extractValues(&ad, data)
if err != nil {
    return err
}
```

TacoVox:
Hi @kortschak!

Thanks for your nice comments! I think I have fixed all the issues that you have outlined.

A comment in regard to naked returns, etc pp.: maybe I could create a P/R in regard to a feasible linter setup, so the Travis check would actually fail if the code does not look a certain way? Of course we would need to agree on a proper set of linters + this would eventually lead to some fixing of the source (which should not be an issue after all).

h5pt_test.go
```diff
@@ -18,20 +18,20 @@ const (
...
-	max_speed   uint64  `hdf5:"Maximum Speed"`
-	longi       int64   `hdf5:"Longitude"`
-	pressure    float32 `hdf5:"Pressure"`
-	temperature float64 `hdf5:"Temperature"`
-	accurate    bool    `hdf5:"Accurate"`
-	// isthep      []int                     // FIXME(sbinet)
-	// jmohep [2][2]int64                    // FIXME(sbinet)
+	// Name        string  `hdf5:"Name"` // FIXME(TacoVox): ReadPackets seems to need an update
+	Vehicle_no  uint8   `hdf5:"Vehicle Number"`
+	Satellites  int8    `hdf5:"Sattelites"`
```
kortschak: Satellites in struct tag too. :)

kortschak:
If there's a linter for naked returns that might be worth running, but it's not a clear cut case of always avoiding them, so incorporating it in the build is probably more trouble than it's worth.

TacoVox:
I think a linting setup in general would improve the code of the whole lib. Maybe just add `golint`? It doesn't have to be `gometalinter` with a full stack of linters :P

TacoVox:
Haha, great that this fails - has nothing to do with the code, I guess?

kortschak:
I've restarted the failing test.

My experience with linting in builds has not been good. Vetting is OK (and now mandatory), but linting involves too much style.

TacoVox:
Hi @sbinet,

added the nil for bot the ad.ptr and each string. I am actually not quite sure that this can happen, as the pointer that is passed to `C.H5PTappend` (in case of appending a single string only) will point to an address that will then contain the address of the string to append.

PR #19, opened 2018-05-23
hdf5: ensure appendData.strs pointers are nil

sbinet:


sbinet:
@TacoVox PTAL

TacoVox:
Ups. I thought if I had to do it that way, but wasn't quite sure and hoped for comments :P Thanks for clarification!

PR #20, opened 2018-05-23
[Experimental] Marshal C Data to []byte

TacoVox:
First and foremost: this is an experimental P/R that works. But the main purpose is for me to get your feedback and maybe we can work something out that is nice and could replace at least the suboptimal `C.realloc` calls.

As @sbinet pointed out in https://github.com/gonum/hdf5/pull/17, it could be nice to marshal data into a `[]byte`. I coulndn't wait, started coding on another branch, and this is the result:
- Replaced the nasty `C.realloc` calls with a simple `make` and `copy` to increase the `[]byte` size.
- Got rid of the `C.free` that was necessary as a result of the memory allocation.
- Experimented a bit and I think I found an acceptable way to find the "native" endianess of the machine. It seems to work as expected.
- Replaced most of the C types with Go-native ones.

What I was not able to achieve is to get rid of the usage of `unsafe` related things. I hope we can find a solution for that. If I am not mistaken, it is obviously a problem to marshal `C.float` and `C.double` into a `uint32` or `uint64` without using `unsafe.Pointer` and a cast. But while experimenting, I found the same for the `unsafe.Pointer` address to the `C.String`s, and the signed integers. A potential solution would have been to extend the `ad.buffer` by 8 bytes before the case statement and then just use `unsafe.Pointer` in those mentioned cases and for the unsigned integers e.g. a `nativeEndian.PutUint16` instead. But for consistency, I liked this solution more.

Furthermore, the strings are as they are. I think, limited by the HDF5 library itself, there is no other way for us to move away from that.

With that: I would be very glad about any input. There is no rush in getting this PR in, I just wanted to exchange thoughts with you guys on that.

If and only if we will get this in, I will change the commit message ;)

### DIFF
--- a/cmd/test-go-table-01/main.go
+++ b/cmd/test-go-table-01/main.go
@@ -75,13 +75,14 @@ func main() {
 	}
 
 	// write one packet to the packet table
-	if err = table.Append(&particles[0]); err != nil {
+	if err = table.Append(particles[0]); err != nil {
 		panic(fmt.Errorf("Append failed with single packet: %s", err))
 	}
 
 	// write several packets
-	parts := particles[1:]
-	if err = table.Append(&parts); err != nil {
+	if err = table.Append(particles[1], particles[2], particles[3],
+		particles[4], particles[5], particles[6], particles[7],
+	); err != nil {
 		panic(fmt.Errorf("Append failed with multiple packets: %s", err))
 	}
 
@@ -109,7 +110,7 @@ func main() {
 
 	// reset index
 	table.CreateIndex()
-	parts = make([]particle, nrecords)
+	parts := make([]particle, nrecords)
 	if err = table.ReadPackets(0, nrecords, &parts); err != nil {
 		panic(fmt.Errorf("ReadPackets failed: %s", err))
 	}
--- a/cmem/cmem.go
+++ b/cmem/cmem.go
@@ -1,0 +1,13 @@
+// Copyright ©2018 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package cmem provides helper functionality for accessing memory in a C
+// compatible fashion.
+package cmem
+
+// CMarshaler is an interface for types that can marshal their data into a
+// C compatible binary layout.
+type CMarshaler interface {
+	MarshalC() ([]byte, error)
+}
--- a/cmem/encoder.go
+++ b/cmem/encoder.go
@@ -1,0 +1,167 @@
+// Copyright ©2018 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package cmem
+
+import (
+	"encoding/binary"
+	"fmt"
+	"math"
+	"reflect"
+	"unsafe"
+)
+
+var nativeEndian binary.ByteOrder
+
+func init() {
+	one := uint16(0x1)
+	nativeEndian = [...]binary.ByteOrder{
+		binary.BigEndian,
+		binary.LittleEndian,
+	}[(*[2]byte)(unsafe.Pointer(&one))[0]]
+}
+
+// Encoder is a wrapper type for information necessary to create and
+// subsequently write an in memory object to a PacketTable using Append().
+type Encoder struct {
+	// Buf contains the encoded data.
+	Buf    []byte
+	offset int
+}
+
+// Encode encodes the value passed as data to binary form stored in []Buf. This
+// buffer is a Go representation of a C in-memory object that can be appended
+// to e.g. a HDF5 PacketTable.
+//
+// Struct values must only have exported fields, otherwise Encode will panic.
+func (enc *Encoder) Encode(data interface{}) error {
+	padding := enc.offset - len(enc.Buf)
+
+	if padding > 0 {
+		enc.Buf = append(enc.Buf, make([]byte, padding)...)
+	}
+
+	if data, ok := data.(CMarshaler); ok {
+		raw, err := data.MarshalC()
+		if err != nil {
+			return err
+		}
+
+		enc.Buf = append(enc.Buf, raw...)
+
+		return nil
+	}
+
+	rv := reflect.Indirect(reflect.ValueOf(data))
+	rt := rv.Type()
+
+	switch rt.Kind() {
+	case reflect.Slice, reflect.Array:
+		for i := 0; i < rv.Len(); i++ {
+			if err := enc.Encode(rv.Index(i).Interface()); err != nil {
+				return err
+			}
+		}
+		return nil
+
+	case reflect.Struct:
+		offset := enc.offset
+		for i := 0; i < rv.NumField(); i++ {
+			sfv := rv.Field(i).Interface()
+			// In order to keep the memory offset always correct, we use the
+			// structs offset.
+			enc.offset = offset + int(rt.Field(i).Offset)
+			if err := enc.Encode(sfv); err != nil {
+				return err
+			}
+			// Reset the offset to the correct array size.
+			enc.offset = offset + int(rt.Size())
+		}
+		return nil
+
+	case reflect.String:
+		str := append([]byte(rv.String()), 0)
+
+		// This direct machine conversion is only used
+		// because HDF5 uses machine endianism.
+		//
+		// DO NOT DO THIS AT HOME.
+		//
+		raw := (*[unsafe.Sizeof(uintptr(0))]byte)(unsafe.Pointer(&str))
+		enc.Buf = append(enc.Buf, raw[:]...)
+		enc.offset += len(raw)
+
+	case reflect.Ptr:
+		return enc.Encode(rv.Elem())
+
+	case reflect.Int8:
+		enc.Buf = append(enc.Buf, byte(rv.Int()))
+		enc.offset++
+
+	case reflect.Uint8:
+		enc.Buf = append(enc.Buf, byte(rv.Uint()))
+		enc.offset++
+
+	case reflect.Int16:
+		var raw [2]byte
+		nativeEndian.PutUint16(raw[:], uint16(rv.Int()))
+		enc.Buf = append(enc.Buf, raw[:]...)
+		enc.offset += 2
+
+	case reflect.Uint16:
+		var raw [2]byte
+		nativeEndian.PutUint16(raw[:], uint16(rv.Uint()))
+		enc.Buf = append(enc.Buf, raw[:]...)
+		enc.offset += 2
+
+	case reflect.Int32:
+		var raw [4]byte
+		nativeEndian.PutUint32(raw[:], uint32(rv.Int()))
+		enc.Buf = append(enc.Buf, raw[:]...)
+		enc.offset += 4
+
+	case reflect.Uint32:
+		var raw [4]byte
+		nativeEndian.PutUint32(raw[:], uint32(rv.Uint()))
+		enc.Buf = append(enc.Buf, raw[:]...)
+		enc.offset += 4
+
+	case reflect.Int64:
+		var raw [8]byte
+		nativeEndian.PutUint64(raw[:], uint64(rv.Int()))
+		enc.Buf = append(enc.Buf, raw[:]...)
+		enc.offset += 8
+
+	case reflect.Uint64:
+		var raw [8]byte
+		nativeEndian.PutUint64(raw[:], uint64(rv.Uint()))
+		enc.Buf = append(enc.Buf, raw[:]...)
+		enc.offset += 8
+
+	case reflect.Float32:
+		var raw [4]byte
+		nativeEndian.PutUint32(raw[:], math.Float32bits(float32(rv.Float())))
+		enc.Buf = append(enc.Buf, raw[:]...)
+		enc.offset += 4
+
+	case reflect.Float64:
+		var raw [8]byte
+		nativeEndian.PutUint64(raw[:], math.Float64bits(rv.Float()))
+		enc.Buf = append(enc.Buf, raw[:]...)
+		enc.offset += 8
+
+	case reflect.Bool:
+		val := byte(0)
+		if rv.Bool() {
+			val = 1
+		}
+		enc.Buf = append(enc.Buf, val)
+		enc.offset++
+
+	default:
+		return fmt.Errorf("hdf5: PT Append does not support datatype (%s)", rt.Kind())
+	}
+
+	return nil
+}
--- a/cmem/encoder_test.go
+++ b/cmem/encoder_test.go
@@ -1,0 +1,56 @@
+// Copyright ©2018 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package cmem
+
+import (
+	"bytes"
+	"encoding/binary"
+	"strconv"
+	"testing"
+)
+
+func TestEncode(t *testing.T) {
+	for i, tc := range []struct {
+		v    interface{}
+		want []byte
+	}{
+		{
+			v: struct {
+				V1 uint8
+				V2 uint64
+				V3 uint8
+				V4 uint16
+			}{1, 2, 3, 4},
+			want: []byte{
+				0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, // |........|
+				0x02, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, // |........|
+				0x03, 0x00, 0x04, 0x00, /*                   */ // |....|
+			},
+		},
+		{
+			v: []int32{1, 20, 300, 4000, 50000, 600000, 7000000, 8000000},
+			want: []byte{
+				0x01, 0x00, 0x00, 0x00, 0x14, 0x00, 0x00, 0x00, // |........|
+				0x2c, 0x01, 0x00, 0x00, 0xa0, 0x0f, 0x00, 0x00, // |,.......|
+				0x50, 0xc3, 0x00, 0x00, 0xc0, 0x27, 0x09, 0x00, // |P....'..|
+				0xc0, 0xcf, 0x6a, 0x00, 0x00, 0x12, 0x7a, 0x00, // |..j...z.|
+			},
+		},
+	} {
+		t.Run(strconv.Itoa(i), func(t *testing.T) {
+			oldEndian := nativeEndian
+			nativeEndian = binary.LittleEndian
+			defer func() { oldEndian = nativeEndian }()
+			var enc Encoder
+			err := enc.Encode(tc.v)
+			if err != nil {
+				t.Fatalf("could not encode: %v", err)
+			}
+			if !bytes.Equal(enc.Buf, tc.want) {
+				t.Fatalf("encoding error:\ngot = %v\nwant= %v", enc.Buf, tc.want)
+			}
+		})
+	}
+}
--- a/h5pt.go
+++ b/h5pt.go
@@ -16,27 +16,9 @@ import (
 	"reflect"
 	"runtime"
 	"unsafe"
+
+	"gonum.org/v1/hdf5/cmem"
 )
-
-// appendData is a wrapper type for information necessary to create and
-// subsequently write an in memory object to a PacketTable using Append().
-type appendData struct {
-	ptr        unsafe.Pointer
-	memSize    C.size_t
-	offset     C.size_t
-	numRecords C.size_t
-	strs       []unsafe.Pointer
-}
-
-func (ad *appendData) free() {
-	C.free(ad.ptr)
-	ad.ptr = nil
-
-	for i, str := range ad.strs {
-		C.free(str)
-		ad.strs[i] = nil
-	}
-}
 
 // Table is an hdf5 packet-table.
 type Table struct {
@@ -102,147 +84,26 @@ func (t *Table) ReadPackets(start, nrecords int, data interface{}) error {
 	return h5err(err)
 }
 
-// extractValues extracts the values to be appended to a packet table and adds
-// them to the appendData structure.
-//
-// Struct values must only have exported fields, otherwise extractValues will
-// panic.
-func extractValues(ad *appendData, data interface{}) error {
-	rv := reflect.Indirect(reflect.ValueOf(data))
-	rt := rv.Type()
-
-	var dataPtr unsafe.Pointer
-	var dataSize C.size_t
-
-	switch rt.Kind() {
-	case reflect.Slice, reflect.Array:
-		for i := 0; i < rv.Len(); i++ {
-			if err := extractValues(ad, rv.Index(i).Interface()); err != nil {
-				return err
-			}
-		}
-		ad.numRecords = C.size_t(rv.Len())
-		return nil
-
-	case reflect.Struct:
-		offset := ad.offset
-		for i := 0; i < rv.NumField(); i++ {
-			sfv := rv.Field(i).Interface()
-			// In order to keep the struct offset always correct
-			ad.offset = offset + C.size_t(rt.Field(i).Offset)
-			if err := extractValues(ad, sfv); err != nil {
-				return err
-			}
-			// Reset the offset to the correct array sized
-			ad.offset = offset + C.size_t(rt.Size())
-		}
-		ad.numRecords = 1
-		return nil
-
-	case reflect.String:
-		ad.numRecords = 1
-		stringData := C.CString(rv.String())
-		ad.strs = append(ad.strs, unsafe.Pointer(stringData))
-		dataPtr = unsafe.Pointer(&stringData)
-		dataSize = C.size_t(unsafe.Sizeof(dataPtr))
-
-	case reflect.Ptr:
-		ad.numRecords = 1
-		return extractValues(ad, rv.Elem())
-
-	case reflect.Int8:
-		ad.numRecords = 1
-		val := C.int8_t(rv.Int())
-		dataPtr = unsafe.Pointer(&val)
-		dataSize = 1
-
-	case reflect.Uint8:
-		ad.numRecords = 1
-		val := C.uint8_t(rv.Uint())
-		dataPtr = unsafe.Pointer(&val)
-		dataSize = 1
-
-	case reflect.Int16:
-		ad.numRecords = 1
-		val := C.int16_t(rv.Int())
-		dataPtr = unsafe.Pointer(&val)
-		dataSize = 2
-
-	case reflect.Uint16:
-		ad.numRecords = 1
-		val := C.uint16_t(rv.Uint())
-		dataPtr = unsafe.Pointer(&val)
-		dataSize = 2
-
-	case reflect.Int32:
-		ad.numRecords = 1
-		val := C.int32_t(rv.Int())
-		dataPtr = unsafe.Pointer(&val)
-		dataSize = 4
-
-	case reflect.Uint32:
-		ad.numRecords = 1
-		val := C.uint32_t(rv.Uint())
-		dataPtr = unsafe.Pointer(&val)
-		dataSize = 4
-
-	case reflect.Int64:
-		ad.numRecords = 1
-		val := C.int64_t(rv.Int())
-		dataPtr = unsafe.Pointer(&val)
-		dataSize = 8
-
-	case reflect.Uint64:
-		ad.numRecords = 1
-		val := C.uint64_t(rv.Uint())
-		dataPtr = unsafe.Pointer(&val)
-		dataSize = 8
-
-	case reflect.Float32:
-		ad.numRecords = 1
-		val := C.float(rv.Float())
-		dataPtr = unsafe.Pointer(&val)
-		dataSize = 4
-
-	case reflect.Float64:
-		ad.numRecords = 1
-		val := C.double(rv.Float())
-		dataPtr = unsafe.Pointer(&val)
-		dataSize = 8
-
-	case reflect.Bool:
-		ad.numRecords = 1
-		val := C.uchar(0)
-		if rv.Bool() {
-			val = 1
-		}
-		dataPtr = unsafe.Pointer(&val)
-		dataSize = 1
-
-	default:
-		return fmt.Errorf("hdf5: PT Append does not support datatype (%s).", rt.Kind())
-	}
-
-	ad.memSize = dataSize + ad.offset
-	ad.ptr = C.realloc(ad.ptr, ad.memSize)
-	C.memcpy(unsafe.Pointer(uintptr(ad.ptr)+uintptr(ad.offset)), dataPtr, dataSize)
-	ad.offset += dataSize
-
-	return nil
-}
-
 // Append appends packets to the end of a packet table.
 //
 // Struct values must only have exported fields, otherwise Append will panic.
-func (t *Table) Append(data interface{}) error {
-	var ad appendData
-	defer ad.free()
-
-	if err := extractValues(&ad, data); err != nil {
-		return err
+func (t *Table) Append(args ...interface{}) error {
+	if len(args) == 0 {
+		return fmt.Errorf("hdf5: no arguments passed to packet table append.")
 	}
 
-	return h5err(C.H5PTappend(t.id, ad.numRecords, ad.ptr))
+	var enc cmem.Encoder
+	for _, arg := range args {
+		if err := enc.Encode(arg); err != nil {
+			return err
+		}
+	}
+
+	if len(enc.Buf) <= 0 {
+		return fmt.Errorf("hdf5: invalid empty buffer")
+	}
+
+	return h5err(C.H5PTappend(t.id, C.size_t(len(args)), unsafe.Pointer(&enc.Buf[0])))
 }
 
 // Next reads packets from a packet table starting at the current index into the value pointed at by data.
--- a/h5pt_test.go
+++ b/h5pt_test.go
@@ -34,7 +34,7 @@ type particle struct {
 	// jmohep      [2][2]int64 // FIXME(sbinet)
 }
 
-func testTable(t *testing.T, dType interface{}, data interface{}, nrecords int) {
+func testTable(t *testing.T, dType interface{}, data ...interface{}) {
 	var table *Table
 
 	typeString := reflect.TypeOf(data).String()
@@ -58,7 +58,7 @@ func testTable(t *testing.T, dType interface{}, data interface{}, nrecords int) 
 	}
 
 	// write one packet to the packet table
-	if err = table.Append(data); err != nil {
+	if err = table.Append(data...); err != nil {
 		t.Fatalf("Append failed with single packet for %s: %s", typeString, err)
 	}
 
@@ -67,8 +67,8 @@ func testTable(t *testing.T, dType interface{}, data interface{}, nrecords int) 
 	if err != nil {
 		t.Fatalf("NumPackets failed for %s: %s", typeString, err)
 	}
-	if n != nrecords {
-		t.Fatalf("Wrong number of packets reported for %s, expected %d but got %d", typeString, nrecords, n)
+	if n != len(data) {
+		t.Fatalf("Wrong number of packets reported for %s, expected %d but got %d", typeString, len(data), n)
 	}
 
 	// iterate through packets
@@ -87,9 +87,9 @@ func testTable(t *testing.T, dType interface{}, data interface{}, nrecords int) 
 	// reset index
 	table.CreateIndex()
 
-	readdata := make([]particle, nrecords)
+	readdata := make([]particle, len(data))
 
-	if err = table.ReadPackets(0, nrecords, &readdata); err != nil {
+	if err = table.ReadPackets(0, len(data), &readdata); err != nil {
 		t.Fatalf("ReadPackets failed for %s: %s", typeString, err)
 	}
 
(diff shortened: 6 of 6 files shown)


sbinet:
that's indeed something along the same lines I was thinking about.
but with a twist:

```diff
diff --git a/h5pt.go b/h5pt.go
index 6576759..cc681ec 100644
--- a/h5pt.go
+++ b/h5pt.go
@@ -12,12 +12,16 @@ package hdf5
 import "C"
 
 import (
+	"encoding/binary"
 	"fmt"
+	"math"
 	"reflect"
 	"runtime"
 	"unsafe"
 )
 
+var native binary.ByteOrder = binary.LittleEndian
+
 // appendData is a wrapper type for information necessary to create and
 // subsequently write an in memory object to a PacketTable using Append().
 type appendData struct {
@@ -26,6 +30,8 @@ type appendData struct {
 	offset     C.size_t
 	numRecords C.size_t
 	strs       []unsafe.Pointer
+
+	cbuf []byte
 }
 
 func (ad *appendData) free() {
@@ -108,6 +114,15 @@ func (t *Table) ReadPackets(start, nrecords int, data interface{}) error {
 // Struct values must only have exported fields, otherwise extractValues will
 // panic.
 func extractValues(ad *appendData, data interface{}) error {
+	if data, ok := data.(interface{ CMarshal() ([]byte, error) }); ok {
+		raw, err := data.CMarshal()
+		if err != nil {
+			return err
+		}
+		ad.cbuf = append(ad.cbuf, raw...)
+		return nil
+	}
+
 	rv := reflect.Indirect(reflect.ValueOf(data))
 	rt := rv.Type()
 
@@ -145,6 +160,16 @@ func extractValues(ad *appendData, data interface{}) error {
 		ad.strs = append(ad.strs, unsafe.Pointer(stringData))
 		dataPtr = unsafe.Pointer(&stringData)
 		dataSize = C.size_t(unsafe.Sizeof(dataPtr))
+		switch unsafe.Sizeof(uintptr(0)) {
+		case 4:
+			var raw [4]byte
+			native.PutUint32(raw[:], uint32(uintptr(dataPtr)))
+			ad.cbuf = append(ad.cbuf, raw[:]...)
+		case 8:
+			var raw [8]byte
+			native.PutUint64(raw[:], uint64(uintptr(dataPtr)))
+			ad.cbuf = append(ad.cbuf, raw[:]...)
+		}
 
 	case reflect.Ptr:
 		ad.numRecords = 1
@@ -155,60 +180,86 @@ func extractValues(ad *appendData, data interface{}) error {
 		val := C.int8_t(rv.Int())
 		dataPtr = unsafe.Pointer(&val)
 		dataSize = 1
+		ad.cbuf = append(ad.cbuf, byte(rv.Int()))
 
 	case reflect.Uint8:
 		ad.numRecords = 1
 		val := C.uint8_t(rv.Uint())
 		dataPtr = unsafe.Pointer(&val)
 		dataSize = 1
+		ad.cbuf = append(ad.cbuf, byte(rv.Uint()))
 
 	case reflect.Int16:
 		ad.numRecords = 1
 		val := C.int16_t(rv.Int())
 		dataPtr = unsafe.Pointer(&val)
 		dataSize = 2
+		var raw [2]byte
+		native.PutUint16(raw[:], uint16(rv.Int()))
+		ad.cbuf = append(ad.cbuf, raw[:]...)
 
 	case reflect.Uint16:
 		ad.numRecords = 1
 		val := C.uint16_t(rv.Uint())
 		dataPtr = unsafe.Pointer(&val)
 		dataSize = 2
+		var raw [2]byte
+		native.PutUint16(raw[:], uint16(rv.Uint()))
+		ad.cbuf = append(ad.cbuf, raw[:]...)
 
 	case reflect.Int32:
 		ad.numRecords = 1
 		val := C.int32_t(rv.Int())
 		dataPtr = unsafe.Pointer(&val)
 		dataSize = 4
+		var raw [4]byte
+		native.PutUint32(raw[:], uint32(rv.Int()))
+		ad.cbuf = append(ad.cbuf, raw[:]...)
 
 	case reflect.Uint32:
 		ad.numRecords = 1
 		val := C.uint32_t(rv.Uint())
 		dataPtr = unsafe.Pointer(&val)
 		dataSize = 4
+		var raw [4]byte
+		native.PutUint32(raw[:], uint32(rv.Uint()))
+		ad.cbuf = append(ad.cbuf, raw[:]...)
 
 	case reflect.Int64:
 		ad.numRecords = 1
 		val := C.int64_t(rv.Int())
 		dataPtr = unsafe.Pointer(&val)
 		dataSize = 8
+		var raw [8]byte
+		native.PutUint64(raw[:], uint64(rv.Int()))
+		ad.cbuf = append(ad.cbuf, raw[:]...)
 
 	case reflect.Uint64:
 		ad.numRecords = 1
 		val := C.uint64_t(rv.Uint())
 		dataPtr = unsafe.Pointer(&val)
 		dataSize = 8
+		var raw [8]byte
+		native.PutUint64(raw[:], uint64(rv.Uint()))
+		ad.cbuf = append(ad.cbuf, raw[:]...)
 
 	case reflect.Float32:
 		ad.numRecords = 1
 		val := C.float(rv.Float())
 		dataPtr = unsafe.Pointer(&val)
 		dataSize = 4
+		var raw [4]byte
+		native.PutUint32(raw[:], math.Float32bits(float32(rv.Float())))
+		ad.cbuf = append(ad.cbuf, raw[:]...)
 
 	case reflect.Float64:
 		ad.numRecords = 1
 		val := C.double(rv.Float())
 		dataPtr = unsafe.Pointer(&val)
 		dataSize = 8
+		var raw [8]byte
+		native.PutUint64(raw[:], math.Float64bits(rv.Float()))
+		ad.cbuf = append(ad.cbuf, raw[:]...)
 
 	case reflect.Bool:
 		ad.numRecords = 1
@@ -218,6 +269,7 @@ func extractValues(ad *appendData, data interface{}) error {
 		}
 		dataPtr = unsafe.Pointer(&val)
 		dataSize = 1
+		ad.cbuf = append(ad.cbuf, uint8(val))
 
 	default:
 		return fmt.Errorf("hdf5: PT Append does not support datatype (%s).", rt.Kind())
```

sbinet:
ignore the shortcut I took wrt `native`.
also: a better handling of the alignment of fields should be applied. (I mean: implemented, really)

then, one would just need to pass `unsafe.Pointer(&ad.cbuf[0])` to the `C.h5pt` function.
handling of strings is still cumbersome (and I guess the would-be interface `CMarshal` couldn't actually work w/o passing also some reference to the `appendData` or at least way to register pointers to `*C.char`...)

TacoVox:
That looks really nice. I can try to integrate your proposal into my code so far and test it. Of course you are also free to change my code, as you should be able in here. We just need to sync :sunglasses: 

* Yes, the alignment is a must, if structs have to be supported (they are a very valuable in my opinion).
* Yes, the complicated string handling won't go away, as we are using the variable length ones. And as far as I understand the HDF5 documentation, the library expects a pointer to the memory containing the variable length string. So the not-so-nice way of keeping the reference in `appendData` would need to stay in, as well as the obvious call to `C.free`.

sbinet:
feel free to rip off my code.
it was meant that way :)

TacoVox:
I ripped of your code, but I added you as co-author ;)

So, as you can see, still some modifications from your original source, as we have to ensure the offset. Otherwise, I think the code speaks for itself :P

h5pt.go
```diff
@@ -153,102 +170,87 @@ func extractValues(ad *appendData, data interface{}) error {
...
-		val := C.double(rv.Float())
-		dataPtr = unsafe.Pointer(&val)
-		dataSize = 8
+		nativeEndian.PutUint64(ad.cBuffer[ad.offset:],
+			math.Float64bits(rv.Float()))
+		ad.offset += 8
 
 	case reflect.Bool:
 		ad.numRecords = 1
 		val := C.uchar(0)
```
sbinet: just use `byte` directly instead of `C.uchar` ?

h5pt.go
```diff
@@ -118,11 +119,27 @@ func (t *Table) ReadPackets(start, nrecords int, data interface{}) error {
...
+		return nil
+	}
+
 	rv := reflect.Indirect(reflect.ValueOf(data))
 	rt := rv.Type()
 
-	var dataPtr unsafe.Pointer
-	var dataSize int
+	// We will at most append 8 bytes at one time.
+	tmp := make([]byte, ad.offset+8)
```
sbinet: I am not sure what we try to do here...
is this done to ensure we have at least `ad.offset+8` of capacity ?

sbinet: ah. ok. now I get it.

a perhaps clearer way (at least to me, YMMV) to ensure we have the correct "padding" (or alignment) would be:
```go
ad.cbuf = append(ad.cbuf, make([]byte, padding)...)
```
where padding is some value we correctly compute :)

TacoVox: Yes. It's to ensure that the padding is correct.

Yeah your way would make that cleaner. Maybe I will wrap it in an if-statement so that we just run it whenever `padding > 0` and `padding := ad.offset - len(ad.cBuffer)` .

Will fix tmrw.

h5pt.go
```diff
@@ -118,11 +119,27 @@ func (t *Table) ReadPackets(start, nrecords int, data interface{}) error {
...
 func extractValues(ad *appendData, data interface{}) error {
+	if data, ok := data.(interface{ CMarshal() ([]byte, error) }); ok {
+		raw, err := data.CMarshal()
+		if err != nil {
+			return err
+		}
+
+		tmp := make([]byte, ad.offset+len(raw))
+		copy(tmp, ad.cBuffer)
+		tmp = ad.cBuffer
```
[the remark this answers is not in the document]
TacoVox: That is actually the wrong way around. I have no idea, why the test cases passed 😆 
Should obviously be `ad.cBuffer = tmp`. But with the proposed change, this will fall away.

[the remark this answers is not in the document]
sbinet: probably b/c nobody's implementing this made-up interface :)

h5pt.go
sbinet: SGTM.

h5pt.go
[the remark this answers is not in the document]
TacoVox: Haha thought this is the coder further below xD Just making noobish comments one after another. Should just shut up xD

h5pt.go
```diff
@@ -12,26 +12,37 @@ package hdf5
...
 	"fmt"
+	"math"
 	"reflect"
 	"runtime"
 	"unsafe"
 )
 
+var nativeEndian binary.ByteOrder
+
+func init() {
```
kortschak:
```
var (
    one uint16 = 1
    nativeEndian = [...]binary.ByteOrder{
        binary.BigEndian,
        binary.LittleEndian,
    }[(*[2]byte)(unsafe.Pointer(&one))[0]]
)
```

I'll see myself out (you can write Perl in any language).

h5pt.go
```diff
@@ -12,26 +12,37 @@ package hdf5
...
+	val := uint16(0x1)
+	valPtr := (*[2]byte)(unsafe.Pointer(&val))
+
+	if valPtr[0] == 1 {
```
kortschak: However, I'd just put the expression here and index it.

h5pt.go
```diff
@@ -108,11 +119,27 @@ func (t *Table) ReadPackets(start, nrecords int, data interface{}) error {
 // Struct values must only have exported fields, otherwise extractValues will
 // panic.
 func extractValues(ad *appendData, data interface{}) error {
+	if data, ok := data.(interface{ CMarshal() ([]byte, error) }); ok {
```
kortschak: It's nicer to name interfaces that get used like this, just for the sake of readability.

h5pt.go
```diff
@@ -12,26 +12,37 @@ package hdf5
...
+}
+
 // appendData is a wrapper type for information necessary to create and
 // subsequently write an in memory object to a PacketTable using Append().
 type appendData struct {
-	ptr        unsafe.Pointer
-	memSize    C.size_t
-	offset     C.size_t
-	numRecords C.size_t
+	cBuffer    []byte
```
kortschak: I think we could have a name that's more informative than `cBuffer`. It's a buffer of bytes, so it can just be called `buf`. The `c` actually detracts from the name unless it means something other than character, in which case it's not informative enough and should be spelled out.

h5pt.go
```diff
@@ -108,11 +119,27 @@ func (t *Table) ReadPackets(start, nrecords int, data interface{}) error {
...
+
+		return nil
+	}
+
 	rv := reflect.Indirect(reflect.ValueOf(data))
 	rt := rv.Type()
 
-	var dataPtr unsafe.Pointer
-	var dataSize C.size_t
+	// We will at most append 8 bytes at one time.
```
kortschak: This doesn't seem like it's always needed (and when it is, not necessarily this much, so we run ahead of ourselves). In the first case of the big switch, e.g., there's no addition to the buffer. Can we not just use `append` at the site of appending?

h5pt.go
```diff
@@ -143,91 +170,87 @@ func extractValues(ad *appendData, data interface{}) error {
...
 	case reflect.Bool:
 		ad.numRecords = 1
 		val := C.uchar(0)
 		if rv.Bool() {
 			val = 1
 		}
-		dataPtr = unsafe.Pointer(&val)
-		dataSize = 1
+		ad.cBuffer[ad.offset] = byte(val)
+		ad.offset += 1
```
kortschak: `ad.offset++`

h5pt.go
TacoVox: Of course! WIP :)

h5pt.go
TacoVox: Will be gone with @sbinet propsal.

TacoVox:
According to my `go test`, we are hitting lines 127-129 with our current test. So we have some unalignment in there.

sbinet:
good :)
I am wondering whether we shouldn't also add a test for 32b alignment...

h5pt.go
```diff
@@ -12,32 +12,46 @@ package hdf5
...
+}
+
 // appendData is a wrapper type for information necessary to create and
 // subsequently write an in memory object to a PacketTable using Append().
 type appendData struct {
-	ptr        unsafe.Pointer
-	memSize    C.size_t
-	offset     C.size_t
-	numRecords C.size_t
+	buf        []byte
```
sbinet: it dawns on me that now, `appendData` looks more and more like an encoder, pretty much like `encoding/json.Encoder`.

`extractValues` could be a method on that `encoder` type:
```go
func (enc *encoder) encode(v interface{}) error { ... }
```

h5pt.go
```diff
@@ -143,91 +171,105 @@ func extractValues(ad *appendData, data interface{}) error {
 		ad.numRecords = 1
 		stringData := C.CString(rv.String())
 		ad.strs = append(ad.strs, unsafe.Pointer(stringData))
-		dataPtr = unsafe.Pointer(&stringData)
-		dataSize = C.size_t(unsafe.Sizeof(dataPtr))
+		dataPtr := uintptr(unsafe.Pointer(&stringData))
```
sbinet: I am wondering whether we couldn't get away with:

```go
str := append([]byte(rv.String()), 0)
ad.strs = append(ad.strs, unsafe.Pointer(&str[0]))
dataPtr := uintptr(ad.strs[len(ad.strs)-1])
```

then we don't have to worry about `C.free` anymore...

TacoVox: It seems to work.
But we do not need to append it to the ad.strs any longer. Patch incoming!

TacoVox:
These type of changes are exactly what I wanted to achieve with this collaboration PR! Pretty awesome. I think that the `C.String` call was a big weakness of the whole thing.

TacoVox:
As you can see, I moved to a method of that `encoder` type. That makes sense to me and looks IMHO  good.
The only thing that I am not really fond of is the support of a `CMarhsaller` interface. Why would we like to support that? Is that not causing more issues that it could solve for the end-user?

sbinet:
`CMarshal` would allow to bypass (slow) reflection.

the code could even be automatically generated by a `go/types`-based program, like I did here:
- https://godoc.org/go-hep.org/x/hep/brio/cmd/brio-gen
- https://godoc.org/go-hep.org/x/hep/brio/cmd/brio-gen/internal/gen
- https://github.com/go-hep/hep/blob/master/brio/cmd/brio-gen/internal/gen/gen.go#L78

TacoVox:
Now that is an argument that I can buy. After all those context switches to C are gone, `reflect` will be our biggest bottleneck besides the IO itself.
Maybe we should add this in the `CMarshal` interface documentation. I will make an attempt in that regard.

TacoVox:
Obviously, I am open for rewording.

Plus, I would be interested in your opinion regarding organization of the files.  Right now, I added the methods under the corresponding structs, but maybe you want to have a different order there? First interfaces, structs then methods?

sbinet:
I think I would personally move the encoder stuff into its own file (and even perhaps its own package).
I'll go through a deeper PR this afternoon (Paris time).

h5pt.go
```diff
@@ -12,237 +12,274 @@ package hdf5
...
 
-// Table is an hdf5 packet-table.
-type Table struct {
-	Identifier
+// CMarshaller is an interface for types that can marshal their data into an
+// H5PTappend compatible []byte. This can be particularly interesting when used
+// in combination with code generation. Furthermore this is a more efficient way
+// of appending data, as it is more efficient than the reflection-based way of
+// doing so which is used under when unmarshalled data is used with Append().
+type CMarshaller interface {
```
sbinet: to follow the stdlib, I think we should name it `CMarshaler` (one `l`):

```
$> go doc encoding.BinaryMarshaler 
type BinaryMarshaler interface {
	MarshalBinary() (data []byte, err error)
}
    BinaryMarshaler is the interface implemented by an object that can marshal
    itself into a binary form.

    MarshalBinary encodes the receiver into a binary form and returns the
    result.

```

h5pt.go
```diff
@@ -12,237 +12,274 @@ package hdf5
...
+	CMarshal() ([]byte, error)
```
sbinet: and this should be named `MarshalC`

h5pt.go
```diff
@@ -12,237 +12,274 @@ package hdf5
...
+	numRecords int
+	strs       []unsafe.Pointer
 }
 
-func (t *Table) finalizer() {
-	if err := t.Close(); err != nil {
-		panic(fmt.Errorf("error closing packet table: %s", err))
-	}
-}
+// extractValues extracts the values to be appended to a packet table and adds
```
sbinet: the comment doesn't match any more what it's commenting :)

h5pt.go
```diff
@@ -12,237 +12,274 @@ package hdf5
...
+// doing so which is used under when unmarshalled data is used with Append().
+type CMarshaller interface {
+	CMarshal() ([]byte, error)
 }
 
-func newPacketTable(id C.hid_t) *Table {
-	t := &Table{Identifier{id}}
-	runtime.SetFinalizer(t, (*Table).finalizer)
-	return t
+// cEncoder is a wrapper type for information necessary to create and
```
sbinet: `s/cEncoder/encoder/`

h5pt.go
```diff
@@ -12,237 +12,274 @@ package hdf5
...
+// subsequently write an in memory object to a PacketTable using Append().
+type encoder struct {
+	buf        []byte
+	offset     int
+	numRecords int
+	strs       []unsafe.Pointer
```
sbinet: `strs` isn't needed anymore, is it?

h5pt.go
```diff
@@ -12,237 +12,274 @@ package hdf5
...
+	return h5err(err)
+}
+
 // Append appends packets to the end of a packet table.
 //
 // Struct values must only have exported fields, otherwise Append will panic.
 func (t *Table) Append(data interface{}) error {
-	var ad appendData
-	defer ad.free()
+	enc := &encoder{}
```
sbinet: `s/enc := &encoder{}/var enc encoder/`

h5pt.go
```diff
@@ -12,237 +12,274 @@ package hdf5
...
 
-	if err := extractValues(&ad, data); err != nil {
+	if err := enc.encode(data); err != nil {
 		return err
 	}
 
-	return h5err(C.H5PTappend(t.id, ad.numRecords, ad.ptr))
+	cData := unsafe.Pointer(&enc.buf[0])
```
sbinet: we should probably check for the length of that buffer before doing this, at least to send the user a legible or more interesting error, in case that buffer is `nil` or empty.

cmem/encoder.go
```diff
@@ -0,0 +1,168 @@
...
+	"fmt"
+	"math"
+	"reflect"
+	"unsafe"
+)
+
+var nativeEndian binary.ByteOrder
+
+func init() {
+	one := uint16(0x1)
```
kortschak: Sorry, that was really meant to be a joke. It can stay though.

sbinet:
yes, it's tested via `h5pt_test.go` but it's also entangled with whatever `hdf5` does under the hood which may break (our assumptions) at some point in the future.

one could probably add a few simple tests in `encoder_test.go` to test alignment, slices, structs and arrays.
for tests purposes, one could probably add a `binary.ByteOrder` in `Encoder` that would be automatically initialized to `nativeEndian` in `Encode` if `nil`, and initialized to, say, `binary.LittleEndian` in the tests:

```go
type Encoder struct {
    bo     binary.ByteOrder
    Buf    []byte
    Offset int
}

func (enc *Encoder) Encode(data interface{}) error {
    if enc.bo == nil { enc.bo = nativeEndian }
    // use enc.bo instead of nativeEndian directly
    ...
}
```

then, we'd have a stable expected output against which we could write tests.
(hum... except for 32/64b alignment...)

or, do w/o this additional field and, within each test, swap-in/swap-out `nativeEndian` with, say, `binary.LittleEndian`... (good old "switcheroo")

TacoVox:
These two are now *really* simple. But I do not have the creativity today... I guess. :man_dancing:

cmem/encoder_test.go
```diff
@@ -0,0 +1,97 @@
...
+)
+
+type testStruct struct {
+	Val1 uint8
+	Val2 uint16
+	Val3 uint32
+	Val4 uint64
+}
+
+func testCmemStructAlignment(t *testing.T, endianess binary.ByteOrder) {
```
sbinet: I was thinking more along the lines of:
```go
func TestEncode(t *testing.T) {
	for i, tc := range []struct {
		v    interface{}
		want []byte
	}{
		{
			v: struct {
				V1 uint8
				V2 uint64
				V3 uint8
			}{1, 2, 3},
			want: []byte("some binary string"),
		},
		{
			v: []int32{1, 20, 300, 4000, 50000, 600000, 7000000, 8000000},
			want: []byte("some other binary string"),
		},
	} {
		t.Run(strconv.Itoa(i), func(t *testing.T) {
			oldEndian := nativeEndian
			nativeEndian = binary.LittleEndian
			defer func() { oldEndian = nativeEndian }()
			var enc Encoder
			err := enc.Encode(tc.v)
			if err != nil {
				t.Fatalf("could not encode: %v", err)
			}
			if !bytes.Equal(enc.Buf, tc.want) {
				t.Fatalf("encoding error:\ngot = %v\nwant= %v", enc.Buf, tc.want)
			}
		})
	}
}

```

TacoVox: Quite beautiful, let me work with that :)

sbinet:
in retrospect, I think it was a mistake to use finalizers.
I certainly didn't know better at that time :)

but now:
- https://crawshaw.io/blog/tragedy-of-finalizers

I guess this could be extracted as its own issue :)

TacoVox:
> I guess this could be extracted as its own issue :)

:+1: 

Other than that, are the files in `cmd` really necessary? Seems to me that they test for stuff that is tested in the `_test.go` files already?

sbinet:
they could probably be rewritten as proper `ExampleSomething`.
I was a much younger padawan at that time :)

sbinet:
I filed https://github.com/gonum/hdf5/issues/22

sbinet:
sure.

but (last nitpick), the commit messages should be slightly modified to read, respectively:

```
hdf5{,/cmem}: ...
```
and
```
hdf5/cmem: ...
```

(for the first one, one could just do with `hdf5/cmem: ...` as well)

TacoVox:
As you wish :paintbrush:

cmem/cmem.go
```diff
@@ -0,0 +1,16 @@
...
+
+// Package cmem provides helper functionality for accessing memory in a C
+// compatible fashion.
+package cmem
+
+// CMarshaler is an interface for types that can marshal their data into an
+// C compatible binary layout. This can be particularly interesting when used in
+// combination with code generation. Furthermore this is a more efficient way of
+// appending data, as it is more efficient than the reflection-based way of
+// doing so which is used under when unmarshalled data is used with Append().
```
sbinet: not sure the comment about `Append()` makes sense in this new general context.

h5pt.go
```diff
@@ -102,147 +84,27 @@ func (t *Table) ReadPackets(start, nrecords int, data interface{}) error {
...
-// Append appends packets to the end of a packet table.
-//
-// Struct values must only have exported fields, otherwise Append will panic.
-func (t *Table) Append(data interface{}) error {
-	var ad appendData
-	defer ad.free()
-
-	if err := extractValues(&ad, data); err != nil {
-		return err
+	if len(enc.Buf) < 0 {
```
sbinet: `s/</<=/` or just `==`

h5pt.go
```diff
@@ -102,147 +84,27 @@ func (t *Table) ReadPackets(start, nrecords int, data interface{}) error {
...
+		return fmt.Errorf("hdf5: the encoded buffer is empty and can " +
```
sbinet: I feel like the error message is bit longwinded.
I would go with `"hdf5: invalid empty buffer"`

but I am not strongly opiniated about this :)

h5pt.go
TacoVox: Actually... It should be `<=` IMHO. Don't know what I thought here...

cmem/cmem.go
TacoVox: Removed the part about `Append()`. Makes it a bit easier to read now, I guess...

kortschak:
I'll be able to look at this tonight.

cmd/test-go-table-01/main.go
```diff
@@ -75,13 +75,14 @@ func main() {
...
 	// write one packet to the packet table
-	if err = table.Append(&particles[0]); err != nil {
+	if err = table.Append(particles[0]); err != nil {
 		panic(fmt.Errorf("Append failed with single packet: %s", err))
 	}
 
 	// write several packets
-	parts := particles[1:]
-	if err = table.Append(&parts); err != nil {
+	if err = table.Append(particles[1], particles[2], particles[3],
```
kortschak: `err = Append(particles[1:]...)`

Also, I'm not a huge fan of unnecessary wrapping of assignments in if constructs, but this was here, so I guess it can stay.

cmem/cmem.go
```diff
@@ -0,0 +1,14 @@
+// Copyright ©2018 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package cmem provides helper functionality for accessing memory in a C
+// compatible fashion.
+package cmem
+
+// CMarshaler is an interface for types that can marshal their data into an
```
kortschak: s/an/a/

cmem/cmem.go
```diff
@@ -0,0 +1,14 @@
...
+// C compatible binary layout. This can be particularly interesting when used in
```
kortschak: Delete the second sentence here.

cmem/encoder.go
```diff
@@ -0,0 +1,171 @@
...
+	nativeEndian = [...]binary.ByteOrder{
+		binary.BigEndian,
+		binary.LittleEndian,
+	}[(*[2]byte)(unsafe.Pointer(&one))[0]]
+}
+
+// Encoder is a wrapper type for information necessary to create and
+// subsequently write an in memory object to a PacketTable using Append().
+type Encoder struct {
+	Buf    []byte
```
kortschak: Why are these exported? If they need to be, there should be doc comments on them.

cmem/encoder.go
```diff
@@ -0,0 +1,171 @@
...
+				return err
+			}
+		}
+		return nil
+
+	case reflect.Struct:
+		offset := enc.Offset
+		for i := 0; i < rv.NumField(); i++ {
+			sfv := rv.Field(i).Interface()
+			// In order to keep the struct offset always correct
```
kortschak: Can you make this a complete sentence explaining what is is that you're doing to keep the offset always correct.

cmem/encoder.go
```diff
@@ -0,0 +1,171 @@
...
+			enc.Offset = offset + int(rt.Field(i).Offset)
+			if err := enc.Encode(sfv); err != nil {
+				return err
+			}
+			// Reset the offset to the correct array sized
```
kortschak: This should either be "Reset the offset to the correct array size." or "Reset the offset to the correct array sized <something>."

cmem/encoder.go
```diff
@@ -0,0 +1,171 @@
...
+			enc.Offset = offset + int(rt.Size())
+		}
+		return nil
+
+	case reflect.String:
+		str := append([]byte(rv.String()), 0)
+		strPtr := uintptr(unsafe.Pointer(&str[0]))
+		dataSize := int(unsafe.Sizeof(strPtr))
+
+		switch dataSize {
```
kortschak: Interestingly, this can be written

```
var raw [unsafe.Sizeof(strPtr)]byte
switch len(raw) {
case 4:
    nativeEndian.PutUint32(raw[:], uint32(strPtr))
case 8:
    nativeEndian.PutUint64(raw[:], uint64(strPtr))
}

enc.Buf = append(enc.Buf, raw[:]...)
end.Offset += len(raw)
```

`unsafe.Sizeof(T)` is a compiler generated constant, and `len(raw)` is a compile time constant.

cmem/encoder.go
```diff
@@ -0,0 +1,171 @@
...
+		enc.Offset += 8
+
+	case reflect.Uint64:
+		var raw [8]byte
+		nativeEndian.PutUint64(raw[:], uint64(rv.Uint()))
+		enc.Buf = append(enc.Buf, raw[:]...)
+		enc.Offset += 8
+
+	case reflect.Float32:
+		enc.Offset += 4
```
kortschak: Please drop this to below the append to match the other stanzas.

cmem/encoder.go
```diff
@@ -0,0 +1,171 @@
...
+	case reflect.Bool:
+		val := byte(0)
+		if rv.Bool() {
+			val = 1
+		}
+		enc.Buf = append(enc.Buf, val)
+		enc.Offset++
+
+	default:
+		return fmt.Errorf("hdf5: PT Append does not support datatype (%s).", rt.Kind())
```
kortschak: Remove trailing period.

cmem/encoder.go
TacoVox: On the contrary this is not the case for most of the fields in the lib :sunglasses: Will do though.

cmem/encoder.go
TacoVox: Nice one!

cmem/encoder.go
TacoVox: Good point.

cmd/test-go-table-01/main.go
TacoVox: Results in `./main.go:83: cannot use particles[1] (type particle) as type []interface {} in argument to table.Append`

kortschak: Yes, sorry. Can you put this in a loop then?

cmem/encoder.go
```diff
@@ -0,0 +1,170 @@
...
+			// Reset the offset to the correct array size.
+			enc.offset = offset + int(rt.Size())
+		}
+		return nil
+
+	case reflect.String:
+		str := append([]byte(rv.String()), 0)
+		strPtr := uintptr(unsafe.Pointer(&str[0]))
+
+		var raw [unsafe.Sizeof(strPtr)]byte
```
kortschak: As another note. Given that hdf5 "does all the wrong things here", we can actually write this whole thing as:

```
// This direct machine conversion is only used
// because HDF5 uses machine endianism.
//
// DO NOT DO THIS AT HOME.
//
raw := (*[unsafe.Sizeof(strPtr)]byte)(strPtr)
enc.Buf = append(enc.Buf, raw[:]...)
enc.offset += len(raw)
```

This is horrific, but in this context, given that they already are already contributing to the net global pain, this is probably the best approach (the comment should be included verbatim).

kortschak:
@TacoVox can you please leave the commits in the history here. It's difficult to go back to previous thoughts when the history has been rebased away.

cmd/test-go-table-01/main.go
TacoVox: Sure!

TacoVox: But actually: that would defeat the purpose of this test, wont you agree? This is supposed to test to append multiple `particle` at once as opposed to a test like

```
for _, part := range particles {
	if err = table.Append(part); err != nil {
		panic(err)
	}
}
```

which would append one at a time.

cmem/encoder_test.go
```diff
@@ -0,0 +1,50 @@
...
+				V2 uint64
+				V3 uint8
+				V4 uint16
+			}{1, 2, 3, 4},
+			want: []byte{0x1, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x2, 0x0, 0x0,
+				0x0, 0x0, 0x0, 0x0, 0x0, 0x3, 0x0, 0x4, 0x0},
+		},
+		{
+			v: []int32{1, 20, 300, 4000, 50000, 600000, 7000000, 8000000},
+			want: []byte{0x1, 0x0, 0x0, 0x0, 0x14, 0x0, 0x0, 0x0, 0x2c, 0x1,
```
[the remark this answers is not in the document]
TacoVox: It breaks my personal 80c limit in a horrific manner, but I will include it :sunglasses:

TacoVox:
As requested by @kortschak, I made new commits. Needless to say, I will squash them into the previous ones (because they are in two different contexts [test and prod code], I created the two separate commits) before this thang is merged down.

cmd/test-go-table-01/main.go
sbinet: Yes, I'd like to actually test that code path :)

cmem/encoder_test.go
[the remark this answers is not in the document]
kortschak: You still use punch cards? I included a width option for that case - set `utter.Config.BytesWidth = 8` before `utter.Dump`.

cmem/encoder.go
kortschak: Not just the comment, but the three lines below as well. If we're going to be bad, let's do it properly.

cmem/encoder_test.go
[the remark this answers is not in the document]
TacoVox: Yes I do ;)

cmem/encoder.go
TacoVox: Actually... That doesn't work. It will yield in:
`./encoder.go:99: cannot convert strPtr (type uintptr) to type *[8]byte`

cmd/test-go-table-01/main.go
kortschak: In that case, can you do a loop to load a `[]interface{}` and then use `Append(p...)`, and when this becomes a test rather than a main, it shouldn't be an example. It does seem an odd thing to test though - what is the rationale for it?

cmem/encoder.go
kortschak: Offset as well please.

cmem/encoder.go
kortschak: I left some of it out :|

`raw := (*[unsafe.Sizeof(strPtr)]byte)(unsafe.Pointer(strPtr))`

cmem/encoder.go
TacoVox: Is not exported any longer.

cmem/encoder.go
kortschak: And then continued to be incorrect - edited above.

TacoVox: I tried myself the `unsafe.Pointer(strPtr)` version before, which resulted in a seg violation.
Unfortunately, I have to tell you that your version does the same... :disappointed:

kortschak: I've confirmed that this `raw := (*[unsafe.Sizeof(uintptr(0))]byte)(unsafe.Pointer(&str))` works. Sorry for the bad advice previously.

cmem/encoder.go
kortschak: Sorry, big-little dyslexia.

cmem/encoder.go
TacoVox: Yes! It does. No problemos :)

kortschak:
I'm not sure I agree with merging tests independently of new code. These can all go in one commit. In the case of a regression fix, you might argue that the tests should go in first, but that's not the case here, and not what we usually do.

cmd/test-go-table-01/main.go
TacoVox: I don't exactly know about the rationale here, as I haven't written those files. Just fixed them in order to make them work :)

TacoVox: So I suggest that these things are fixed when https://github.com/gonum/hdf5/issues/22 is tackled.

cmem/encoder.go
TacoVox: :sunglasses:

TacoVox:
Sure, I can just make one commit.

kortschak:
It's important because github (and bitbucket) do a woeful job of retaining history during PRs when rebasing has happened to the branch that's the focus of the PR. Gerrit does it better, but we don't use that at this stage.

TacoVox:
But Gerrit is so horrible in other dimensions IMHO...

kortschak:
Yeah. Everything has problems. We use this one right now.

I'll approve now, but I'm not convinced of the variadic add testing - it looks to me like it's testing a language property, but I'm probably missing some subtlety. It should be cleaned up in an example though, because it's currently showing bad practice in an example.

sbinet:
we could - at some point - use https://gerrithub.io
(but this would also involve a different workflow. at the very least we should also make sure the old one would still be supported)

TacoVox:
Alright. Then I will squash.
I agree with you @kortschak, but I really would change that in a separate PR that you guys should do IMHO.

TacoVox:
And in regard to the multiple commits, maybe add something in `.github/PULL_REQUEST_TEMPLATE` that specifically states that each change should be a separate commit. No matter how dumb the commit message is. And then everything can be squashed together later.

kortschak:
Eeek. The rebase that was just done has killed the connection between the comments and the commits.

When the PR is approved, we can get GitHub to to a squash merge (apparently with a rebase). This allows the branch here to be preserved. Very infrequently, it's worth doing subtle interactive rebases, with the understanding that the comments' contexts will be lost. When there's a lot of discussion I try to avoid that.

TacoVox:
Alrighty. Now it is too late for that. But I guess it ok for this time. Next time I will be aware of the fact.

sbinet:
SGTM. :)

kortschak:
I've come to the realisation that irrespective of the behaviour during the PR, when the branch is deleted, GitHub keeps no record of the commits referred to in in comments other that the pitiful 4 lines in the conversation tab. My life is a lie,... and GitHub is woeful as a platform for code review.

Issue #16, opened 2018-04-26
HDF5 version 1.10 support?


ggggggggg:
How likely is 1.10 support? What needs to be done to support 1.10?

sbinet:
What's new in 1.10? :)

ggggggggg:
I want to use the SWMR funtionality, see https://support.hdfgroup.org/HDF5/docNewFeatures/#1100

It only requires a few new functions. I tried pointing at an hdf5.h from 1.10.1_2 and all the tests seem to pass, so it seems like it works. I originally took the comment to mean it would totally fail on 1.10, but that doesn't seem to be the case. 

As I understand to fully support SWMR one needs access to the functions: `H5Dflush`, `H5DOappend`, `H5Drefresh` and `H5Fstart_swmr_write`.

TacoVox:
Will also affect the `H5PT` interface. But the deprecated functions should still work in 1.10.

TacoVox:
I think this library should not break, as I am running hdf5-dev in version 1.10 on my machine and it just works (I enabled deprecated support when I compiled). Then the calls could be gradually updated via PRs?
Guess it's just a matter of updating the travis config to use `bionic` instead of `trusty` :)

PR #21, opened 2018-05-23
hdf5: first stab at hdf5-1.10 support

sbinet:
This CL is a first step towards supporting new features and APIs
of HDF5 v1.10.

Right now, anecdotal evidence shows that HDF5 v1.8 is still massively
deployed so support for 1.10 has to be opt-in, by way of an optional
build tag aptly named "hdf5-v1.10"

Updates gonum/hdf5#16.

Please take a look.


TacoVox:
That looks good. And to confirm it even further. We are running this lib for quite some time now on HDF5 v1.10.1 without any issues.

kortschak:
>IIUC, the way travis works, we can't reliably test gonum/hdf5 against both hdf5-1.8.x and hdf5-1.10.x: we have to pick one.

I don't think this is true. We could use an environment variable to choose which is installed and built against.

h5f_v1.10.go
```diff
@@ -0,0 +1,15 @@
...
+// license that can be found in the LICENSE file.
+
+//+build hdf5_v1.10
+
+package hdf5
+
+// #include "hdf5.h"
+import "C"
+
+// StartSWMRWrite enables SWMR writing mode for this file.
```
kortschak: Is it worth adding this to the pre-1.10 being an error-returning noop?

h5f_v1.10.go
```diff
@@ -0,0 +1,15 @@
+// Copyright ©2018 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//+build hdf5_v1.10
```
kortschak: I think the convention more broadly has become `// +build ...`.

gustafj:
> IIUC, the way travis works, we can't reliably test gonum/hdf5 against both hdf5-1.8.x and hdf5-1.10.x: we have to pick one.

> I don't think this is true. We could use an environment variable to choose which is installed and built against.

You can use the matrix expansion support in Travis to build against different versions.
I cant say how it would be done in the hdf5 case though (don't know enough about the your build setup).

Some references:
https://docs.travis-ci.com/user/build-stages/matrix-expansion/
https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix

Issue #23, opened 2018-06-20
hdf5: finalizers are sometimes called before they should


TacoVox:
Hello @sbinet. It's me again :smile_cat: 

Using this library in production, I noticed some major issues that seem to be caused by the finalizers which are used.

Three out of four times, our application is crashing. I looked at the traces and found an interesting thing:
```
fatal error: unexpected signal during runtime execution
fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0x80 addr=0x0 pc=0xb762bb]

runtime stack:
runtime.throw(0xcd5fec, 0x2a)
        /usr/local/go/src/runtime/panic.go:616 +0x81
runtime.sigpanic()
        /usr/local/go/src/runtime/signal_unix.go:372 +0x28e

goroutine 4 [syscall]:
runtime.cgocall(0x88e8a0, 0xc42006b698, 0x0)
        /usr/local/go/src/runtime/cgocall.go:128 +0x64 fp=0xc42006b658 sp=0xc42006b620 pc=0x401f94
decode/vendor/gonum.org/v1/hdf5._Cfunc_H5Tclose(0x30000000000013e, 0x0)
        _cgo_gotypes.go:913 +0x4d fp=0xc42006b698 sp=0xc42006b658 pc=0x71391d
decode/vendor/gonum.org/v1/hdf5.(*Datatype).Close(0xc42022cbd0, 0xc42006b738, 0x0)
        /go/src/decode/vendor/gonum.org/v1/hdf5/h5t.go:153 +0x37 fp=0xc42006b6d0 sp=0xc42006b698 pc=0x717ac7
decode/vendor/gonum.org/v1/hdf5.(*Datatype).finalizer(0xc42022cbd0)
        /go/src/decode/vendor/gonum.org/v1/hdf5/h5t.go:138 +0x2f fp=0xc42006b728 sp=0xc42006b6d0 pc=0x7179ef
runtime.call32(0x0, 0xce3b48, 0xc420432000, 0x1000000010)
        /usr/local/go/src/runtime/asm_amd64.s:573 +0x3b fp=0xc42006b758 sp=0xc42006b728 pc=0x4535fb
runtime.runfinq()
        /usr/local/go/src/runtime/mfinal.go:222 +0x1eb fp=0xc42006b7e0 sp=0xc42006b758 pc=0x415edb
runtime.goexit()
[...]

PR #24, opened 2018-06-20
hdf5: removes finalizers

TacoVox:
As discussed in https://github.com/gonum/hdf5/issues/23

> As the finalizers are causing tremendous problems in production usage,
they were completely removed from all types that were using them. Therefore it is
required that the caller of the New* functions also take care of calling the Close() methods.

### DIFF
--- a/h5a.go
+++ b/h5a.go
@@ -10,10 +10,7 @@ package hdf5
 import "C"
 
 import (
-	"fmt"
-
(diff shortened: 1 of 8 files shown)


TacoVox:
Closed in favour of https://github.com/gonum/hdf5/pull/33 that fixes conflicts.

PR #25, opened 2018-06-21
cmem: return error for invalid reflect.Value

TacoVox:
Once again: I noticed something fishy that we haven't taken care of. This is an actual problem using the PacketTable append. Instead of handleing the error in a nice way, the whole thing will just panic, as it is now. Therefore, I created this small fix. Hope you guys agree.

> In case reflect.ValueOf results in an invalid value, the code previously
caused a panic, which is not optimal for error handling.
This patch fixes that by returning an error in case reflect.ValueOf
resulted in an invalid value that cannot be used for further operations.

### DIFF
--- a/cmem/encoder.go
+++ b/cmem/encoder.go
@@ -54,6 +54,10 @@ func (enc *Encoder) Encode(data interface{}) error {
 	}
 
 	rv := reflect.Indirect(reflect.ValueOf(data))
+	if !rv.IsValid() {
+		return fmt.Errorf("cmem: reflect.ValueOf returned invalid value for type %T", data)
+	}
+
 	rt := rv.Type()
 
 	switch rt.Kind() {


cmem/encoder.go
```diff
@@ -54,6 +55,10 @@ func (enc *Encoder) Encode(data interface{}) error {
 	}
 
 	rv := reflect.Indirect(reflect.ValueOf(data))
+	if !rv.IsValid() {
+		return errors.New("cmem: reflect.ValueOf returned invalid value")
```
sbinet: perhaps also add `%T` so we know the original input type that failed?

TacoVox:
Good call.

cmem/encoder.go
```diff
@@ -54,6 +54,11 @@ func (enc *Encoder) Encode(data interface{}) error {
 	}
 
 	rv := reflect.Indirect(reflect.ValueOf(data))
+	if !rv.IsValid() {
+		return fmt.Errorf("cmem: reflect.ValueOf returned invalid value for "+
```
sbinet: I wouldn't split that string. ok, it runs a bit long but who cares.
alternatively:
```go
return fmt.Errorf(
    "cmem: reflect.ValueOf returned invalid value for type %T",
    data,
)
```

sbinet:
out of curiosity, what were the kind of data being mishandled by `Encode` ?

cmem/encoder.go
TacoVox: I care... Hehe :P But I can change it to your suggestion.

kortschak: Yeah, please don't do this. We're not using Fortran and we're not on punch cards.

TacoVox:
So what we are appending are generally large structs with structs (or even arrays of structs) as members. These can be up to several kB per `Append()`. It was a rare edge case when this happened the last time, so I cannot really tell you what the type was. Hard to reproduce... 

I can just give you part of the trace that I received from the developer that faced it:
```
panic: reflect: call of reflect.Value.Type on zero Value

goroutine 72 [running]:
reflect.Value.Type(0x0, 0x0, 0x0, 0x0, 0x0)
	/usr/local/go/src/reflect/value.go:1713 +0x16d
decode/vendor/gonum.org/v1/hdf5/cmem.(*Encoder).Encode(0xc42115bb90, 0x0, 0x0, 0x437201, 0x3)
	/go/src/decode/vendor/gonum.org/v1/hdf5/cmem/encoder.go:57 +0x277
decode/vendor/gonum.org/v1/hdf5.(*Table).Append(0xc420c50850, 0xc42115bbf0, 0x1, 0x1, 0x2, 0x2)
	/go/src/decode/vendor/gonum.org/v1/hdf5/h5pt.go:89 +0x96
decode/internal/app/write/hdfwrite.hdf5Table.append(0xc4202685f0, 0xc42020c0a0, 0x12, 0xc420c50850, 0x0, 0x0, 0x0, 0x0)
```

kortschak:
You won't get my blessing for that typography :|

TacoVox:
It's now as fugly as it gets but sure...

kortschak:
On my workstation, with a pair of half maximum terminals I get
```
$ tput cols
105
```

My laptop gives similar results, and I imagine that most other people's do too. Conceptually related components should be grouped within the bounds of reason. Leaving this as one line does that; at some length it becomes unreasonable, but 90 single width runes is not that length.

Also see ref https://github.com/golang/go/commit/a625b919163e76c391f2865d1f956c0f16d90f83.

TacoVox:
Fair point, but as I believe in the all-mighty [Linux Kernel Coding Style](https://www.kernel.org/doc/html/v4.10/process/coding-style.html), I always stick to the 80 char limit and the things in there that I can adapt in Go as well.

I think that we can agree to disagree. It is about personal taste.

kortschak:
The Linux Kernal Coding Style has its place, but that place is populated by C, without a mechanical code formatter and limited code and type static analysis tools. Go has all of those. While I agree that there are good components of that guide, there are also things that I think are abominations, and even if this were not true, using a style guide from another language would be equivalent to using a style guide produced by the Académie Française for written English. For line length, the guide we generally use is whether you can see the important bits when you are in a reasonable editor. If not, and it's not codegen, then it's too long. You'll see this in testing format lines in gonum/gonum where there is a lot to output. We also break lines in BLAS/LAPACK calls to group parameters conceptually. These are not really personal taste, but rather easing readability and grouping related concepts. It's more engineering than taste.

If anything, eschewing written rules in favour of common sense is the approach that's, IMO, best to use.

PR #27, opened 2018-06-23
hdf5: remove finalizers from types that are components of returned types

### LINKED ISSUE #23: hdf5: finalizers are sometimes called before they should

kortschak:
This retains finalizers and so reduces the risk of memory leaks, but removes them from `Datatype` values that are managed as part of compound, array or varlen types.

Please take a look.

Fixes #23.

/cc @TacoVox

### DIFF
--- a/h5t.go
+++ b/h5t.go
@@ -211,6 +211,7 @@ func NewArrayType(base_type *Datatype, dims []int) (*ArrayType, error) {
 	if err := checkID(hid); err != nil {
 		return nil, err
 	}
+	runtime.SetFinalizer(base_type, nil)
 	t := &ArrayType{Datatype{Identifier{hid}}}
 	runtime.SetFinalizer(t, (*ArrayType).finalizer)
 	return t, nil
@@ -249,6 +250,7 @@ func NewVarLenType(base_type *Datatype) (*VarLenType, error) {
 	if err := checkID(id); err != nil {
 		return nil, err
 	}
+	runtime.SetFinalizer(base_type, nil)
 	t := &VarLenType{Datatype{Identifier{id}}}
 	runtime.SetFinalizer(t, (*VarLenType).finalizer)
 	return t, nil
@@ -322,6 +324,7 @@ func (t *CompoundType) MemberType(mbr_idx int) (*Datatype, error) {
 func (t *CompoundType) Insert(name string, offset int, field *Datatype) error {
 	cname := C.CString(name)
 	defer C.free(unsafe.Pointer(cname))
+	runtime.SetFinalizer(field, nil)
 	return h5err(C.H5Tinsert(t.id, cname, C.size_t(offset), field.id))
 }
 


kortschak:
Given https://github.com/gonum/hdf5/issues/23#issuecomment-399967316, this should probably not have been approved.

Closing in favour of #24 (though note conflict - I'm happy to reproduce the intent of that PR in a new one since I did the damage).

PR #33, opened 2018-06-26
hdf5: remove finalizers and document requirement for closing values

### LINKED ISSUE #23: hdf5: finalizers are sometimes called before they should

kortschak:
Please take a look.

Fixes #23.

### DIFF
--- a/h5f.go
+++ b/h5f.go
@@ -12,7 +12,6 @@ import "C"
 
 import (
 	"fmt"
-	"runtime"
 	"unsafe"
 )
 
@@ -40,17 +39,8 @@ type File struct {
 	CommonFG
 }
 
-func (f *File) finalizer() {
-	err := f.closeWith(h5fclose)
-	if err != nil {
-		panic(fmt.Errorf("error closing file: %s", err))
-	}
-}
-
 func newFile(id C.hid_t) *File {
-	f := &File{CommonFG{Identifier{id}}}
-	runtime.SetFinalizer(f, (*File).finalizer)
-	return f
+	return &File{CommonFG{Identifier{id}}}
 }
 
 // Creates an HDF5 file.
@@ -66,7 +56,8 @@ func CreateFile(name string, flags int) (*File, error) {
 	return newFile(hid), nil
 }
 
-// Opens an existing HDF5 file.
+// Open opens and returns an an existing HDF5 file. The returned
+// file must be closed by the user when it is no longer needed.
 func OpenFile(name string, flags int) (*File, error) {
 	c_name := C.CString(name)
 	defer C.free(unsafe.Pointer(c_name))
@@ -79,7 +70,8 @@ func OpenFile(name string, flags int) (*File, error) {
 	return newFile(hid), nil
 }
 
-// Returns a new identifier for a previously-opened HDF5 file.
+// ReOpen returns a new identifier for a previously-opened HDF5 file.
+// The returned file must be closed by the user when it is no longer needed.
 func (f *File) ReOpen() (*File, error) {
 	hid := C.H5Freopen(f.id)
 	if err := checkID(hid); err != nil {
@@ -96,9 +88,8 @@ func IsHDF5(name string) bool {
 	return C.H5Fis_hdf5(c_name) > 0
 }
 
-// Terminates access to an HDF5 file.
+// Close closes the file.
 func (f *File) Close() error {
-	runtime.SetFinalizer(f, nil)
 	return f.closeWith(h5fclose)
 }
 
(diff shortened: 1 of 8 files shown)


h5f.go
```diff
@@ -66,7 +56,8 @@ func CreateFile(name string, flags int) (*File, error) {
 	return newFile(hid), nil
 }
 
-// Opens an existing HDF5 file.
+// Open opens and returns an an existing HDF5 file. The returned
```
TacoVox: Just a general thought: should we maybe stay consistent with the method/function comments and start them with method/function name. Right now it seems to be the case for half of the ones below.

h5s.go
```diff
@@ -73,6 +64,7 @@ func h5sclose(id C.hid_t) C.herr_t {
 }
 
 // CreateSimpleDataspace creates a new simple dataspace and opens it for access.
+//  The returned dataspace must be closed by the user when it is no longer needed.
```
TacoVox: Remove one white space :sunglasses:

h5f.go
kortschak: Yeah, I missed some. The intention was to do "Func does ....". We can fix more later. There's other lint I want to remove elsewhere, so I'm fine with leaving this and repolishing.

h5s.go
kortschak: This one needs to be fixed. Thanks.

h5f.go
TacoVox: SGTM

h5t.go
```diff
@@ -109,13 +109,12 @@ func OpenDatatype(c CommonFG, name string, tapl_id int) (*Datatype, error) {
...
 func NewDatatype(id C.hid_t) *Datatype {
 	t := &Datatype{Identifier{id}}
-	runtime.SetFinalizer(t, (*Datatype).finalizer)
 	return t
 }
 
-// CreateDatatype creates a new datatype.
-// class must be T_COMPUND, T_OPAQUE, T_ENUM or T_STRING.
-// size is the size of the new datatype in bytes.
+// CreateDatatype creates a new datatype. The value of class must be T_COMPUND,
```
sbinet: actually, it's `s/T_COMPUND/T_COMPOUND/` :)

h5t.go
```diff
@@ -264,15 +255,15 @@ type CompoundType struct {
 	Datatype
 }
 
-// NewCompoundType creates a new CompoundType.
-// size is the size in bytes of the compound datatype.
+// NewCompoundType creates a new CompoundType. The/ size is the size in bytes of
```
sbinet: `s|The/|The|`

h5t.go
kortschak: Fixed

h5t.go
kortschak: Fixed

Issue #12, opened 2017-08-12
hdf5: figure out how to resolve the Go pointer to Go pointer issue


kortschak:
The `GODEBUG=cgocheck=0` incantation is untenable in the long run. So we need to sort out how to deal with types holding pointers.

It is not clear to me how to deal with this using the current approach to calling HDF5 functions where entire values are simply handed off to the library. But anything else involves significant reengineering/reimplementation of the HDF5 code in Go.

Note that the example in the tests that fails this runtime check is already incorrect, since the string field that causes the failure is not correctly reoundtripped anyway.

sbinet:
for [gopy](https://github.com/go-python/gopy), I opted for the following:

- when passing a Go value to python, just send an identifier (an `int32` or `int64`) `idtype`
- the identifier is kept in a `map[idtype]unsafe.Pointer` registry so Go won't change the address of the pointer (a reverse map is also built)
- when one needs to call a function on this handle, send that call to the Go side (`idtype` + what call to perform)

kortschak:
In gopy, you still required `GODEBUG=cgocheck=0` from the README.

I do not see how this approach fixes the problem though. The runtime will call the cgo pointer check at some point, and any pointer hiding just pushes the problem into harder to debug areas. The issue that raised this is the `WriteSubset` function which by necessity has Go pointers inside a pointer value, unless we prohibit the serialisation of structs containing slices and strings (and any other types allowed by HDF5 that are reference-like types).

@ianlancetaylor, do you have any suggestions for how to deal with this kind of a problem. The relevant type is [`s1Type`](https://github.com/gonum/hdf5/blob/58b0e5b9994e1f7b836fc129f72062535c0d83e3/cmd/test-go-cpxcmpd/main.go#L23-L29) which is handled [here](https://github.com/gonum/hdf5/blob/master/h5d.go#L125-L155) where it's interpreted into an unsafe.Pointer at L145, but containing a `string` field with its data pointer.

sbinet:
yes, `gopy` `master` is still requiring `GODEBUG=cgocheck=0`.
I am working on this new scheme I devised above.

ianlancetaylor:
@kortschak If Python does not examine the contents of the Go value, and doesn't do anything with it other than pass it back to Go, then it seems to me that the scheme that @sbinet outlines will work.  You can see a roughly similar scheme at https://github.com/swig/swig/blob/master/Lib/go/goruntime.swg#L418 , used to pass Go values through C++ code.

[...]

PR #35, opened 2018-06-29
hdf5: mark types with pointers

kortschak:
This is part of the marking needed to know whether data copying will be needed. Note that it currently is overly conservative, and for all practical uses it will mark everything. There is a TODO that has an approach that I think will fix that.

This is more a PR for discussion of approach, but may end up being something that we can merge.

The half of the story needed to handle a `Datatype` that has been returned from opening a file is not dealt with here.

Please take a look.

Updates #12.


h5t_types.go
```diff
@@ -152,10 +153,14 @@ func (t *Datatype) Committed() bool {
 	return C.H5Tcommitted(t.id) > 0
 }
 
-// Copy copies an existing datatype. The returned datatype must be closed by the
-// user when it is no longer needed.
+// Copy copies an existing datatype.
 func (t *Datatype) Copy() (*Datatype, error) {
-	return copyDatatype(t.id)
+	c, err := copyDatatype(t.id)
```
sbinet: perhaps modify the signature of `copyDatatype` to take a `*Datatype` instead of `C.hid_t` so nobody would unadvertantly drop the `_hasGoPointer` value on the floor?

kortschak: Sure, I'll do that when I make `NewDatatype` do the same thing with `Identifier` (and check where there are others that may also benefit.

h5t_types.go
```diff
@@ -462,7 +471,13 @@ func NewDataTypeFromType(t reflect.Type) (*Datatype, error) {
 		dt = &cdt.Datatype
 
 	case reflect.Ptr:
-		return NewDataTypeFromType(t.Elem())
+		dt, err = NewDataTypeFromType(t.Elem())
```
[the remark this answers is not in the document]
kortschak: Do you want to file an issue for that?

kortschak:
Any comments about pointer path counting?

kortschak:
I have added precise pointer chain accounting.

PTAL

h5t_types.go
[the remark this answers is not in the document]
TacoVox: Sounds reasonable.

h5t_types_test.go
```diff
@@ -4,35 +4,50 @@
...
+		{v: (*uint16)(nil), illegalPtr: false},
+		{v: (*uint32)(nil), illegalPtr: false},
+		{v: (*uint64)(nil), illegalPtr: false},
+		{v: (*float32)(nil), illegalPtr: false},
+		{v: (*float64)(nil), illegalPtr: false},
+		{v: (*string)(nil), illegalPtr: true},
+		{v: (*bool)(nil), illegalPtr: false},
 	}
 
-	for test := range tests {
```
kortschak: This was testing `int`, test is the first var in the range statement.

h5t_types_test.go
```diff
@@ -4,35 +4,50 @@
...
-		NewDatatypeFromValue(test)
-		// Test again for usage with ptrs
-		NewDatatypeFromValue(&test)
```
kortschak: This would have been (but wasn't) testing `*interface{}` which we do not handle.

h5t_types_test.go
```diff
@@ -4,35 +4,50 @@
...
+	for _, test := range tests {
```
TacoVox: Well spotted!!!

h5t_types_test.go
```diff
@@ -4,35 +4,50 @@
...
+		dt, _ := NewDatatypeFromValue(test.v)
```
TacoVox: Any reason to *not* check for the error here?

h5t_types.go
```diff
@@ -17,6 +17,8 @@ import (
 
 type Datatype struct {
 	Identifier
+
+	_goPtrPathLen int
```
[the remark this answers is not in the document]
kortschak: Here it's OK to change to the w/o. Previously I preferred it the other way. I'll make the change.

h5t_types_test.go
kortschak: I was trying to figure out why my test was not working and shoved in a print %T and only got integers.

h5t_types_test.go
kortschak: We'll panic later if we don't succeed. I'll change it.

kortschak:
PTAL

h5t_types.go
```diff
@@ -438,14 +444,18 @@ func NewDataTypeFromType(t reflect.Type) (*Datatype, error) {
 		if err != nil {
 			return nil, err
 		}
+		var ptrPathLen int
 		n := t.NumField()
 		for i := 0; i < n; i++ {
 			f := t.Field(i)
-			var field_dt *Datatype = nil
+			var field_dt *Datatype
```
TacoVox: I know you didn't wrote this in the first place, but maybe change to `fieldDt`? Maybe I should also open an issue for moving to Go-style Camel Case?

kortschak: There are a collection of things like this that I'd like to fix up. Please do file that issue.

Issue #37, opened 2018-07-02
all: move to Camel Case naming


TacoVox:
As discussed e.g. in https://github.com/gonum/hdf5/pull/35, we should move to the Go-way of using Camel Case, as described in [Effective Go](https://golang.org/doc/effective_go.html#mixed-caps).

Right now, we are dealing with a mixture of both Camel Case and more C_style variable names.

PR #38, opened 2018-07-08
all: moves to camel case naming

donkahlero:
Hi guys,
I had a stab at https://github.com/gonum/hdf5/issues/37 and renamed the variables. Took me a bit longer than expected as so many variables are affected and maybe I did not even catch them all...

Nevertheless, some points are still up for discussion and I hope we can discuss them as part of this PR:

**What shall we do about the Datatype vs. DataType issue?** I saw that @sbinet had once [a stab at this](https://github.com/gonum/hdf5/issues/8). I still suggest that we move what is called `Datatype` to `DataType` as part of this PR and then rethink later if we want to move to what you said in https://github.com/gonum/hdf5/pull/9, @kortschak. The same goes for `Dataspace` which should maybe also be called `DataSpace`?

**Id vs. ID**. Right now I moved everything to `Id`, but I am not quite sure if `ID` is the better way. The go standard lib uses the prior while `golint` suggest the latter. Any opinion?

**_go_types in h5t_types.go**. Should we rename e.g. `_go_int8_t` to `goInt8T`? 

**ALL_CAPS_VARS** as we have them in `h5t_shim.go` should stay as they are, right?

### DIFF
--- a/cmd/test-go-cpxcmpd/main.go
+++ b/cmd/test-go-cpxcmpd/main.go
@@ -67,20 +67,20 @@ func main() {
 		panic(err)
 	}
 	defer f.Close()
-	fmt.Printf(":: file [%s] created (id=%d)\n", fname, f.Id())
+	fmt.Printf(":: file [%s] created (id=%d)\n", fname, f.ID())
 
 	// create the memory data type
-	dType, err := hdf5.NewDatatypeFromValue(s1[0])
+	typ, err := hdf5.NewDatatypeFromValue(s1[0])
 	if err != nil {
-		panic("could not create a dType")
+		panic("could not create a typ")
 	}
 
 	// create the dataset
-	dset, err := f.CreateDataset(dsname, dType, space)
+	dset, err := f.CreateDataset(dsname, typ, space)
 	if err != nil {
 		panic(err)
 	}
-	fmt.Printf(":: dset (id=%d)\n", dset.Id())
+	fmt.Printf(":: dset (id=%d)\n", dset.ID())
 
 	// write data to the dataset
 	fmt.Printf(":: dset.Write...\n")
--- a/h5a_attribute.go
+++ b/h5a_attribute.go
@@ -22,11 +22,11 @@ func newAttribute(id C.hid_t) *Attribute {
 	return &Attribute{Identifier{id}}
 }
 
-func createAttribute(id C.hid_t, name string, dType *Datatype, dSpace *Dataspace, acpl *PropList) (*Attribute, error) {
+func createAttribute(id C.hid_t, name string, typ *Datatype, spc *Dataspace, acpl *PropList) (*Attribute, error) {
 	cName := C.CString(name)
 	defer C.free(unsafe.Pointer(cName))
-	hid := C.H5Acreate2(id, cName, dType.id, dSpace.id, acpl.id, P_DEFAULT.id)
-	if err := checkId(hid); err != nil {
+	hid := C.H5Acreate2(id, cName, typ.id, spc.id, acpl.id, P_DEFAULT.id)
+	if err := checkID(hid); err != nil {
 		return nil, err
 	}
 	return newAttribute(hid), nil
@@ -37,7 +37,7 @@ func openAttribute(id C.hid_t, name string) (*Attribute, error) {
 	defer C.free(unsafe.Pointer(cName))
 
 	hid := C.H5Aopen(id, cName, P_DEFAULT.id)
-	if err := checkId(hid); err != nil {
+	if err := checkID(hid); err != nil {
 		return nil, err
 	}
 	return newAttribute(hid), nil
@@ -68,7 +68,7 @@ func (s *Attribute) Space() *Dataspace {
 }
 
 // Read reads raw data from a attribute into a buffer.
-func (s *Attribute) Read(data interface{}, dType *Datatype) error {
+func (s *Attribute) Read(data interface{}, typ *Datatype) error {
 	var addr unsafe.Pointer
 	v := reflect.ValueOf(data)
 
@@ -88,13 +88,13 @@ func (s *Attribute) Read(data interface{}, dType *Datatype) error {
 		addr = unsafe.Pointer(v.UnsafeAddr())
 	}
 
-	rc := C.H5Aread(s.id, dType.id, addr)
+	rc := C.H5Aread(s.id, typ.id, addr)
 	err := h5err(rc)
 	return err
 }
 
 // Write writes raw data from a buffer to an attribute.
-func (s *Attribute) Write(data interface{}, dType *Datatype) error {
+func (s *Attribute) Write(data interface{}, typ *Datatype) error {
 	var addr unsafe.Pointer
 	v := reflect.Indirect(reflect.ValueOf(data))
 	switch v.Kind() {
@@ -114,7 +114,7 @@ func (s *Attribute) Write(data interface{}, dType *Datatype) error {
 		addr = unsafe.Pointer(v.UnsafeAddr())
 	}
 
-	rc := C.H5Awrite(s.id, dType.id, addr)
+	rc := C.H5Awrite(s.id, typ.id, addr)
 	err := h5err(rc)
 	return err
 }
--- a/h5d_dataset.go
+++ b/h5d_dataset.go
@@ -26,18 +26,18 @@ func newDataset(id C.hid_t, typ *Datatype) *Dataset {
 	return &Dataset{Identifier: Identifier{id}, typ: typ}
 }
 
-func createDataset(id C.hid_t, name string, dType *Datatype, dSpace *Dataspace, dcpl *PropList) (*Dataset, error) {
-	dType, err := dType.Copy() // For safety
+func createDataset(id C.hid_t, name string, typ *Datatype, spc *Dataspace, dcpl *PropList) (*Dataset, error) {
+	typ, err := typ.Copy() // For safety
 	if err != nil {
 		return nil, err
 	}
 	cName := C.CString(name)
 	defer C.free(unsafe.Pointer(cName))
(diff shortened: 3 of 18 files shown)


kortschak:
[This](https://github.com/golang/go/wiki/CodeReviewComments) will answer some of your questions.

Data Type is two words, so it can be made `DataType`, though do think `Type` is better.

Dataspace is a single compound word, so it should remain (English is weird).

ID should be `ID`  - as described in the wiki.

For the `reflect.Type` values, s/_go_(.*)_t/\1Type/g.

It seems reasonable to keep the `T_` vars and consts as they are.

cmd/test-go-cpxcmpd/main.go
```diff
@@ -67,20 +67,20 @@ func main() {
 		panic(err)
 	}
 	defer f.Close()
-	fmt.Printf(":: file [%s] created (id=%d)\n", fname, f.ID())
+	fmt.Printf(":: file [%s] created (id=%d)\n", fname, f.Id())
```
kortschak: Please revert the ID changes.

cmd/test-go-cpxcmpd/main.go
```diff
@@ -67,20 +67,20 @@ func main() {
...
 
 	// create the memory data type
-	dtype, err := hdf5.NewDatatypeFromValue(s1[0])
+	dType, err := hdf5.NewDatatypeFromValue(s1[0])
```
kortschak: I'd be really happy if these all just became `typ`.

h5a_attribute.go
```diff
@@ -22,22 +22,22 @@ func newAttribute(id C.hid_t) *Attribute {
 	return &Attribute{Identifier{id}}
 }
 
-func createAttribute(id C.hid_t, name string, dtype *Datatype, dspace *Dataspace, acpl *PropList) (*Attribute, error) {
-	c_name := C.CString(name)
-	defer C.free(unsafe.Pointer(c_name))
-	hid := C.H5Acreate2(id, c_name, dtype.id, dspace.id, acpl.id, P_DEFAULT.id)
-	if err := checkID(hid); err != nil {
+func createAttribute(id C.hid_t, name string, dType *Datatype, dSpace *Dataspace, acpl *PropList) (*Attribute, error) {
```
kortschak: And, for the `Dataspace` labels, just `spc`.

h5d_dataset.go
```diff
@@ -89,14 +89,14 @@ func (s *Dataset) ReadSubset(data interface{}, memspace, filespace *Dataspace) e
 		addr = unsafe.Pointer(v.UnsafeAddr())
 	}
 
-	var filespace_id, memspace_id C.hid_t = 0, 0
-	if memspace != nil {
-		memspace_id = memspace.id
+	var fileSpaceId, memSpaceId C.hid_t = 0, 0
```
kortschak: Remove the redundant assignment and change to `fileSpaceID, memSpaceID`.

h5d_dataset.go
```diff
@@ -59,9 +59,9 @@ func (s *Dataset) Space() *Dataspace {
 }
 
 // ReadSubset reads a subset of raw data from a dataset into a buffer.
-func (s *Dataset) ReadSubset(data interface{}, memspace, filespace *Dataspace) error {
-	dtype, err := s.Datatype()
-	defer dtype.Close()
+func (s *Dataset) ReadSubset(data interface{}, memSpace, fileSpace *Dataspace) error {
```
kortschak: These are good though.

h5d_dataset.go
```diff
@@ -137,14 +137,14 @@ func (s *Dataset) WriteSubset(data interface{}, memspace, filespace *Dataspace)
 		addr = unsafe.Pointer(v.UnsafeAddr())
 	}
 
-	var filespace_id, memspace_id C.hid_t = 0, 0
-	if memspace != nil {
-		memspace_id = memspace.id
+	var fileSpaceId, memSpaceId C.hid_t = 0, 0
```
kortschak: Same here.

h5d_dataset.go
```diff
@@ -176,11 +176,11 @@ func (s *Dataset) OpenAttribute(name string) (*Attribute, error) {
 // Datatype returns the HDF5 Datatype of the Dataset. The returned
 // datatype must be closed by the user when it is no longer needed.
 func (s *Dataset) Datatype() (*Datatype, error) {
-	dtype_id := C.H5Dget_type(s.id)
-	if dtype_id < 0 {
+	dTypeId := C.H5Dget_type(s.id)
+	if dTypeId < 0 {
```
kortschak: `typID`

h5f_file.go
```diff
@@ -106,36 +106,36 @@ func (f *File) Flush(scope Scope) error {
 // FIXME
 // Retrieves name of file to which object belongs.
 func (f *File) FileName() string {
-	// ssize_t H5Fget_name(hid_t obj_id, char *name, size_t size )
+	// size_t H5Fget_name(hid_t obj_id, char *name, size_t size )
```
kortschak: I can't remember why these are here. I think the signature comments can go away.

h5i_identifier.go
```diff
@@ -44,13 +44,13 @@ func (i Identifier) Name() string {
 		return ""
 	}
 	buf := string(make([]byte, sz))
-	c_buf := C.CString(buf)
-	defer C.free(unsafe.Pointer(c_buf))
-	sz = int(C.H5Iget_name(i.id, c_buf, C.size_t(sz)))
+	cBuf := C.CString(buf)
```
kortschak: Why is this not just `C.malloc(sz+1)`?

h5pt_table.go
```diff
@@ -43,30 +43,30 @@ func (t *Table) IsValid() bool {
...
-func (t *Table) ReadPackets(start, nrecords int, data interface{}) error {
-	c_start := C.hsize_t(start)
-	c_nrecords := C.size_t(nrecords)
+func (t *Table) ReadPackets(start, nRecords int, data interface{}) error {
+	cStart := C.hsize_t(start)
+	cNRecords := C.size_t(nRecords)
 	rv := reflect.Indirect(reflect.ValueOf(data))
 	rt := rv.Type()
-	c_data := unsafe.Pointer(nil)
+	cData := unsafe.Pointer(nil)
```
kortschak: `var cData unsafe.Pointer`

h5pt_table.go
```diff
@@ -129,9 +129,9 @@ func (t *Table) Next(data interface{}) error {
 
 // NumPackets returns the number of packets in a packet table.
 func (t *Table) NumPackets() (int, error) {
-	c_nrecords := C.hsize_t(0)
-	err := C.H5PTget_num_packets(t.id, &c_nrecords)
-	return int(c_nrecords), h5err(err)
+	cNRecords := C.hsize_t(0)
```
kortschak: `var n C.hsize_t`

h5pt_table.go
```diff
@@ -142,57 +142,57 @@ func (t *Table) CreateIndex() error {
 
 // SetIndex sets a packet table's index.
 func (t *Table) SetIndex(index int) error {
-	c_idx := C.hsize_t(index)
-	err := C.H5PTset_index(t.id, c_idx)
+	cIdX := C.hsize_t(index)
```
kortschak: `cIdx`

h5t_types.go
```diff
@@ -440,36 +440,36 @@ func NewDataTypeFromType(t reflect.Type) (*Datatype, error) {
...
 		if err != nil {
 			return nil, err
 		}
 		var ptrPathLen int
 		n := t.NumField()
 		for i := 0; i < n; i++ {
 			f := t.Field(i)
-			var field_dt *Datatype
-			field_dt, err = NewDataTypeFromType(f.Type)
+			var fieldDT *Datatype
```
kortschak: `ft`

h5t_types.go
```diff
@@ -440,36 +440,36 @@ func NewDataTypeFromType(t reflect.Type) (*Datatype, error) {
 
 	case reflect.Struct:
 		sz := int(t.Size())
-		cdt, err := NewCompoundType(sz)
+		cDT, err := NewCompoundType(sz)
```
kortschak: I'd leave this one.

h5t_types_test.go
```diff
@@ -110,30 +110,30 @@ func TestStructDatatype(t *testing.T) {
 
 	// Test that the type can be constructed and that the number of
 	// members is as expected.
-	var dtypes []*Datatype
+	var dTypes []*Datatype
```
kortschak: `types`

h5t_types_test.go
```diff
@@ -110,30 +110,30 @@ func TestStructDatatype(t *testing.T) {
...
 
 	// "Regular" value
-	dtype, err := NewDatatypeFromValue(test)
+	dType, err := NewDatatypeFromValue(test)
```
kortschak: `typ`

hdf5.go
```diff
@@ -16,8 +16,8 @@ import (
 func init() {
 	err := h5err(C.H5open())
 	if err != nil {
-		err_str := fmt.Sprintf("pb calling H5open(): %s", err)
-		panic(err_str)
+		errStr := fmt.Sprintf("pb calling H5open(): %s", err)
```
kortschak: Merge these lines and save the transient label.

h5i_identifier.go
TacoVox: Hmm. Then we have to start casting `sz + 1` to `C.size_t` and `cBuf` to `*C.char`. The question is what is more beautiful?

TacoVox:
Good comments, @kortschak! Thx!

I found one more thing that I really dislike. We have some mix-up in the test cases when it comes to `fname` vs. `fName`. I suggest we move everything to the latter?

Furthermore, I suggest that we also rename `dsname` in the test-cases to `dsName`?

PR #48, opened 2019-03-26
Support dataset compression with ZLIB.

yuanqj8191:
Support dataset compression for both pre-defined compression filter ZLIB in HDF5.

### DIFF
--- a/h5p_proplist.go
+++ b/h5p_proplist.go
@@ -10,8 +10,6 @@ package hdf5
 // inline static
 // hid_t _go_hdf5_H5P_DEFAULT() { return H5P_DEFAULT; }
 // hid_t _go_hdf5_H5P_DATASET_CREATE() { return H5P_DATASET_CREATE; }
-// hid_t _go_hdf5_H5_SZIP_EC_OPTION_MASK() { return H5_SZIP_EC_OPTION_MASK; }
-// hid_t _go_hdf5_H5_SZIP_NN_OPTION_MASK() { return H5_SZIP_NN_OPTION_MASK; }
 import "C"
 
 import "unsafe"
@@ -23,10 +21,8 @@ type PropList struct {
 }
 
 var (
-	P_DEFAULT              *PropList = newPropList(C._go_hdf5_H5P_DEFAULT())
-	H5P_DATASET_CREATE     PropType  = PropType(C._go_hdf5_H5P_DATASET_CREATE()) // Properties for dataset creation
-	H5_SZIP_EC_OPTION_MASK           = uint(C._go_hdf5_H5_SZIP_EC_OPTION_MASK()) // Selects entropy coding method.
-	H5_SZIP_NN_OPTION_MASK           = uint(C._go_hdf5_H5_SZIP_NN_OPTION_MASK()) // Selects nearest neighbor coding method.
+	P_DEFAULT          *PropList = newPropList(C._go_hdf5_H5P_DEFAULT())
+	H5P_DATASET_CREATE PropType  = PropType(C._go_hdf5_H5P_DATASET_CREATE()) // Properties for dataset creation
 )
 
 func newPropList(id C.hid_t) *PropList {
@@ -63,15 +59,6 @@ func (p *PropList) SetChunk(ndims uint, dim []uint) error {
 // https://support.hdfgroup.org/HDF5/doc/RM/RM_H5P.html#Property-SetDeflate
 func (p *PropList) SetDeflate(level uint) error {
 	if err := h5err(C.H5Pset_deflate(C.hid_t(p.id), C.uint(level))); err != nil {
-		return err
-	}
-	return nil
-}
-
-// Sets up use of the SZIP compression filter.
-// https://support.hdfgroup.org/HDF5/doc/RM/RM_H5P.html#Property-SetSzip
-func (p *PropList) SetSzip(optionsMask, pixelsPerBlock uint) error {
-	if err := h5err(C.H5Pset_szip(C.hid_t(p.id), C.uint(optionsMask), C.uint(pixelsPerBlock))); err != nil {
 		return err
 	}
 	return nil
--- a/h5p_proplist.go
+++ b/h5p_proplist.go
@@ -55,16 +55,34 @@ func (p *PropList) Close() error {
 
 // SetChunk sets the size of the chunks used to store a chunked layout dataset.
 // https://support.hdfgroup.org/HDF5/doc/RM/RM_H5P.html#Property-SetChunk
-func (p *PropList) SetChunk(dim []uint) error {
-	ndims := len(dim)
+func (p *PropList) SetChunk(dims []uint) error {
+	ndims := len(dims)
 	if ndims <= 0 {
 		return fmt.Errorf("number of dimensions must be same size as the rank of the dataset, but zero received")
 	}
 	c_dim := make([]C.hsize_t, ndims)
-	for i := range dim {
-		c_dim[i] = C.hsize_t(dim[i])
+	for i := range dims {
+		c_dim[i] = C.hsize_t(dims[i])
 	}
 	return h5err(C.H5Pset_chunk(C.hid_t(p.id), C.int(ndims), &c_dim[0]))
+}
+
+// GetChunk retrieves the size of chunks for the raw data of a chunked layout dataset.
+// https://support.hdfgroup.org/HDF5/doc/RM/RM_H5P.html#Property-GetChunk
+func (p *PropList) GetChunk(ndims int) (dims []uint, err error) {
+	if ndims <= 0 {
+		err = fmt.Errorf("number of dimensions must be same size as the rank of the dataset, but nonpositive value received")
+		return
+	}
+	c_dims := make([]C.hsize_t, ndims)
+	if err = h5err(C.H5Pget_chunk(C.hid_t(p.id), C.int(ndims), &c_dims[0])); err != nil {
+		return
+	}
+	dims = make([]uint, ndims)
+	for i := range dims {
+		dims[i] = uint(c_dims[i])
+	}
+	return
 }
 
 // SetDeflate sets deflate (GNU gzip) compression method and compression level.
--- a/h5p_proplist.go
+++ b/h5p_proplist.go
@@ -7,9 +7,8 @@ package hdf5
 // #include "hdf5.h"
 // #include <stdlib.h>
 // #include <string.h>
-// inline static
-// hid_t _go_hdf5_H5P_DEFAULT() { return H5P_DEFAULT; }
-// hid_t _go_hdf5_H5P_DATASET_CREATE() { return H5P_DATASET_CREATE; }
+// static inline hid_t _go_hdf5_H5P_DEFAULT() { return H5P_DEFAULT; }
+// static inline hid_t _go_hdf5_H5P_DATASET_CREATE() { return H5P_DATASET_CREATE; }
 import "C"
 
 import "unsafe"
@@ -21,8 +20,8 @@ type PropList struct {
 }
 
 var (
-	P_DEFAULT          *PropList = newPropList(C._go_hdf5_H5P_DEFAULT())
-	H5P_DATASET_CREATE PropType  = PropType(C._go_hdf5_H5P_DATASET_CREATE()) // Properties for dataset creation
+	P_DEFAULT        *PropList = newPropList(C._go_hdf5_H5P_DEFAULT())
+	P_DATASET_CREATE PropType  = PropType(C._go_hdf5_H5P_DATASET_CREATE()) // Properties for dataset creation
 )
 
 func newPropList(id C.hid_t) *PropList {
@@ -44,18 +43,21 @@ func (p *PropList) Close() error {
 	return p.closeWith(h5pclose)
 }
 
-// Sets the size of the chunks used to store a chunked layout dataset.
+// SetChunk sets the size of the chunks used to store a chunked layout dataset.
 // https://support.hdfgroup.org/HDF5/doc/RM/RM_H5P.html#Property-SetChunk
-func (p *PropList) SetChunk(ndims uint, dim []uint) error {
-	var c_dim *C.hsize_t
-	c_dim = (*C.hsize_t)(unsafe.Pointer(&dim[0]))
+func (p *PropList) SetChunk(dim []uint) error {
+	ndims := len(dim)
+	if ndims <= 0 {
+		return h5err(C.herr_t(-1))
+	}
+	c_dim := (*C.hsize_t)(unsafe.Pointer(&dim[0]))
 	if err := h5err(C.H5Pset_chunk(C.hid_t(p.id), C.int(ndims), c_dim)); err != nil {
 		return err
 	}
 	return nil
 }
 
-// Sets deflate (GNU gzip) compression method and compression level.
+// SetDeflate sets deflate (GNU gzip) compression method and compression level.
 // https://support.hdfgroup.org/HDF5/doc/RM/RM_H5P.html#Property-SetDeflate
 func (p *PropList) SetDeflate(level uint) error {
 	if err := h5err(C.H5Pset_deflate(C.hid_t(p.id), C.uint(level))); err != nil {
--- a/h5p_proplist.go
+++ b/h5p_proplist.go
@@ -75,9 +75,6 @@ func (p *PropList) SetDeflate(level int) error {
 	if level == DefaultCompression {
 		level = 6
 	}
-	if level < 0 {
-		return fmt.Errorf("unsupported compression level: %d", level)
-	}
 	if err := h5err(C.H5Pset_deflate(C.hid_t(p.id), C.uint(level))); err != nil {
 		return err
 	}
--- a/h5p_proplist.go
+++ b/h5p_proplist.go
@@ -14,7 +14,6 @@ import "C"
 import (
 	"compress/zlib"
 	"fmt"
-	"unsafe"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


kortschak:
Do the tests pass for you locally?

yuanqj8191:
Yes. My local HDF5 is v1.10.0. And the build server is v1.8.11. According to HDF5's change log, this break should be caused by the version mismatch.
https://support.hdfgroup.org/ftp/HDF5/current/src/unpacked/release_docs/HISTORY-1_8_0-1_10_0.txt

yuanqj8191:
Do we have a plan to support v1.10.x?

sbinet:
> Do we have a plan to support v1.10.x?

IIRC, we delayed a bit the explicit support of v1.10.x because at the time it wasn't well distributed (_e.g._ on travis.)

if we manage to setup travis to use 1.10.x, then I guess we could explicitly support 1.10.x (and support 1.8.x on a best effort basis.)

h5p_proplist.go
```diff
@@ -9,16 +9,20 @@ package hdf5
 // #include <string.h>
 // inline static
 // hid_t _go_hdf5_H5P_DEFAULT() { return H5P_DEFAULT; }
+// hid_t _go_hdf5_H5P_DATASET_CREATE() { return H5P_DATASET_CREATE; }
```
sbinet: perhaps make it `inline static` as the one above?

h5p_proplist.go
```diff
@@ -40,6 +44,26 @@ func (p *PropList) Close() error {
 	return p.closeWith(h5pclose)
 }
 
+// Sets the size of the chunks used to store a chunked layout dataset.
```
sbinet: please prepend `SetChunk` in that sentence.

h5p_proplist.go
```diff
@@ -9,16 +9,20 @@ package hdf5
...
 type PropType C.hid_t
 
 type PropList struct {
 	Identifier
 }
 
 var (
-	P_DEFAULT *PropList = newPropList(C._go_hdf5_H5P_DEFAULT())
+	P_DEFAULT          *PropList = newPropList(C._go_hdf5_H5P_DEFAULT())
+	H5P_DATASET_CREATE PropType  = PropType(C._go_hdf5_H5P_DATASET_CREATE()) // Properties for dataset creation
```
sbinet: to be consistent, I believe this should be named `P_DATASET_CREATE`.

h5p_proplist.go
```diff
@@ -40,6 +44,26 @@ func (p *PropList) Close() error {
 	return p.closeWith(h5pclose)
 }
 
+// Sets the size of the chunks used to store a chunked layout dataset.
+// https://support.hdfgroup.org/HDF5/doc/RM/RM_H5P.html#Property-SetChunk
+func (p *PropList) SetChunk(ndims uint, dim []uint) error {
+	var c_dim *C.hsize_t
+	c_dim = (*C.hsize_t)(unsafe.Pointer(&dim[0]))
```
sbinet: couldn't we directly initialize `c_dim`? (_ie:_ merge line 50 and 51 into just one line)
also: we should probably properly handle 0-size and `nil` `dim` slices.

h5p_proplist.go
```diff
@@ -40,6 +44,26 @@ func (p *PropList) Close() error {
...
+	if err := h5err(C.H5Pset_chunk(C.hid_t(p.id), C.int(ndims), c_dim)); err != nil {
+		return err
+	}
+	return nil
+}
+
+// Sets deflate (GNU gzip) compression method and compression level.
```
sbinet: please prepend `SetDeflate` to the doc string.

h5p_proplist_test.go
```diff
@@ -0,0 +1,119 @@
+// Copyright ©2017 The Gonum Authors. All rights reserved.
```
sbinet: `s/2017/2019/`

h5p_proplist_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * These test cases are based on the h5_cmprss.c by The HDF Group.
+ * https://support.hdfgroup.org/HDF5/examples/intro.html#c
+ */
+
+func TestDeflate(t *testing.T) {
+	DisplayErrors(true)
+	fn, dsn, dims := "cmprss_deflate.h5", "dset_cmpress", []uint{1000, 1000}
+	defer DisplayErrors(false)
```
sbinet: I would move that `defer` just after the `DisplayErrors(true)` line.

h5p_proplist_test.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+/**
+ * These test cases are based on the h5_cmprss.c by The HDF Group.
+ * https://support.hdfgroup.org/HDF5/examples/intro.html#c
+ */
+
+func TestDeflate(t *testing.T) {
+	DisplayErrors(true)
+	fn, dsn, dims := "cmprss_deflate.h5", "dset_cmpress", []uint{1000, 1000}
```
sbinet: I'd make that a `var ( ... )` block

h5p_proplist_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+	dset, err := f.CreateDatasetWith(dsn, T_NATIVE_DOUBLE, dspace, dclp)
+	if err != nil {
+		return nil, err
+	}
+	defer dset.Close()
+
+	n := dims[0] * dims[1]
+	data := make([]float64, n)
+	for i := uint(0); i < n; i++ {
```
sbinet: I'd replace this with a `for i := range data`

h5p_proplist_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+	data := make([]float64, dims[0]*dims[1])
+	dset.Read(&data[0])
+	return data, nil
+}
+
+func compare(ds0, ds1 []float64) error {
+	n0, n1 := len(ds0), len(ds1)
+	if n0 != n1 {
+		return errors.New("Dimensions not meet")
```
sbinet: I'd replace this with the following error message:
```go
return fmt.Errorf("dimensions mismatch: %d != %d", n0, n1)
```

h5p_proplist_test.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	for i := 0; i < n0; i++ {
+		d := math.Abs(ds0[i] - ds1[i])
+		if d > 1e-7 {
+			return errors.New("Values not meet")
```
sbinet: I'd replace this with the following error message:
```go
return fmt.Errorf("values at index %d differ: %v != %v", i, ds0[i], ds1[i])
```

h5p_proplist_test.go
```diff
@@ -0,0 +1,119 @@
+// Copyright ©2017 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package hdf5
```
sbinet: I'd put this test file under the `package hdf5_test`.
AFAICT, the test doesn't use any internals of the `hdf5` package.

h5p_proplist.go
```diff
@@ -40,6 +44,26 @@ func (p *PropList) Close() error {
 	return p.closeWith(h5pclose)
 }
 
+// Sets the size of the chunks used to store a chunked layout dataset.
+// https://support.hdfgroup.org/HDF5/doc/RM/RM_H5P.html#Property-SetChunk
+func (p *PropList) SetChunk(ndims uint, dim []uint) error {
```
sbinet: couldn't we drop the `ndims` argument?
AFAICT, we don't need it in Go: slices know their length.

h5p_proplist.go
```diff
@@ -40,6 +44,26 @@ func (p *PropList) Close() error {
...
+	c_dim = (*C.hsize_t)(unsafe.Pointer(&dim[0]))
+	if err := h5err(C.H5Pset_chunk(C.hid_t(p.id), C.int(ndims), c_dim)); err != nil {
+		return err
+	}
+	return nil
+}
+
+// Sets deflate (GNU gzip) compression method and compression level.
+// https://support.hdfgroup.org/HDF5/doc/RM/RM_H5P.html#Property-SetDeflate
+func (p *PropList) SetDeflate(level uint) error {
```
sbinet: perhaps we could make `level` an `int` instead, and piggy-back on the behaviour of `compress/flate` that uses `-1` to mean `DefaultCompression`?

h5p_proplist.go
yuanqj8191: Fixed

h5p_proplist.go
yuanqj8191: Fixed

h5p_proplist.go
yuanqj8191: Fixed

h5p_proplist.go
yuanqj8191: Fixed

h5p_proplist.go
yuanqj8191: Fixed

h5p_proplist_test.go
```diff
@@ -107,12 +112,12 @@ func load(fn, dsn string) ([]float64, error) {
 func compare(ds0, ds1 []float64) error {
 	n0, n1 := len(ds0), len(ds1)
 	if n0 != n1 {
-		return errors.New("Dimensions not meet")
+		return errors.New(fmt.Sprintf("dimensions mismatch: %d != %d", n0, n1))
```
sbinet: Perhaps just use 'fmt.Errorf' directly?

h5p_proplist_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+func compare(ds0, ds1 []float64) error {
+	n0, n1 := len(ds0), len(ds1)
+	if n0 != n1 {
+		return errors.New(fmt.Sprintf("dimensions mismatch: %d != %d", n0, n1))
+	}
+	for i := 0; i < n0; i++ {
+		d := math.Abs(ds0[i] - ds1[i])
+		if d > 1e-7 {
+			return errors.New(fmt.Sprintf("values at index %d differ: %f != %f", i, ds0[i], ds1[i]))
```
sbinet: Ditto.

h5p_proplist_test.go
yuanqj8191: All files are with same license title. Shall we change it all with another PR?

h5p_proplist_test.go
yuanqj8191: Fixed

h5p_proplist_test.go
yuanqj8191: Fixed

h5p_proplist_test.go
yuanqj8191: Fixed

h5p_proplist_test.go
yuanqj8191: Fixed

h5p_proplist_test.go
yuanqj8191: Fixed

h5p_proplist_test.go
yuanqj8191: All test files are put in hdf5 package. Shall we change all of them?

h5p_proplist.go
yuanqj8191: Fixed

h5p_proplist.go
yuanqj8191: Maybe we should keep the behavior in C API to use uint.

h5p_proplist_test.go
sbinet: New files get their copyright year from the year they are written.
We don't update old files' copyright.

h5p_proplist_test.go
sbinet: New files get their copyright year from the year they are written.
We don't update old files' copyright.

h5p_proplist_test.go
yuanqj8191: Fixed

h5p_proplist_test.go
yuanqj8191: Fixed

h5p_proplist_test.go
sbinet: It's more of a case by case basis.
(That said, there are perhaps test files that could be put in a hdf5_test pkg. But thus can be part of another PR)

h5p_proplist_test.go
yuanqj8191: OK. Fixed

h5p_proplist.go
sbinet: it's not really changing its behaviour: it's adding an additional possible value (-1) to mean something that isn't possible to express currently.

`cgo` based packages are a bit akward because one wants to follow the original C API (so switchers from C are in known territory) while at the same provide the Go idioms that made those switchers switch to Go.

anyways, I'll wait for @kortschak to break the tie :)

kortschak: This should be an `int`.

yuanqj8191: I still stick to my opinion. `level` is between 0 and 9, and with no default value.

kortschak: Can you explain why it is that the C trumps Go idiom here?

yuanqj8191: Sorry, I cannot find the rule for this.

kortschak: I don't understand. What rule?

yuanqj8191: Can we communicate more efficiently? Just tell me your reason. 
Here, `level` is always a `uint`, and no need to define a default value. We need extra unnecessary check if use `int`. This is my point.

kortschak: There are a variety of reasons.

1. The compression routines already in use in the standard library use `int` types.
2. There already needs to be a check since any type in Go is able to hold more than the members of the set {0, 1, 2, 3, 4, 5, 6, 7, 8, 9}.
3. The check is not expensive in comparison to all of the other work that is being done by any non-trivial program (and likely even any trivial program).
4. The default integer type is `int`.

I don't see any reason to keep this value as a `uint` except that the code here claims that C uses a `uint` (it doesn't because depending on the platform and compiler `uint` and `unsigned long long` may be different sizes - also note that that makes [this](https://github.com/gonum/hdf5/pull/48/files#diff-96202169867e203414c9ede99d46f5caR48) and [this](https://github.com/gonum/hdf5/pull/48/files#diff-96202169867e203414c9ede99d46f5caR53) unsafe).

The upshot is that hdf5 here is Go code, with Go coding conventions. It wraps C, but does that so that the users of this package do not need to interact directly with the C.

h5p_proplist.go
```diff
@@ -40,6 +43,29 @@ func (p *PropList) Close() error {
...
 }
 
+// SetChunk sets the size of the chunks used to store a chunked layout dataset.
+// https://support.hdfgroup.org/HDF5/doc/RM/RM_H5P.html#Property-SetChunk
+func (p *PropList) SetChunk(dim []uint) error {
+	ndims := len(dim)
+	if ndims <= 0 {
+		return h5err(C.herr_t(-1))
+	}
+	c_dim := (*C.hsize_t)(unsafe.Pointer(&dim[0]))
```
kortschak: This is not safe.

h5p_proplist.go
yuanqj8191: The first one of your reasons is a good point. But the other three are unrelated. I already update it.

h5p_proplist.go
yuanqj8191: Do you have a better solution?

kortschak: A for loop.

h5p_proplist.go
kortschak: No. They are very much related; in fact 4. is essentially a restatement of 1. Number 2. and 3. is a refutation of the importance of "We need extra unnecessary check if use `int`."

h5p_proplist.go
```diff
@@ -58,8 +69,15 @@ func (p *PropList) SetChunk(dim []uint) error {
...
 
 // SetDeflate sets deflate (GNU gzip) compression method and compression level.
+// If level is set as DefaultCompression, 6 will be used.
 // https://support.hdfgroup.org/HDF5/doc/RM/RM_H5P.html#Property-SetDeflate
-func (p *PropList) SetDeflate(level uint) error {
+func (p *PropList) SetDeflate(level int) error {
+	if level == DefaultCompression {
+		level = 6
+	}
+	if level < 0 {
```
kortschak: Check for grater than 9 as well.

sbinet: Or BestCompression.

yuanqj8191: It will be checked in C.

kortschak: If that's the case, there is no need to check whether `level < 0`.

yuanqj8191: The `level` in C is `uint`. A minus value will be meaningless after converted to `uint`. In this case, Go API user cannot understand the returning error from C.

kortschak: This is not true. You are performing a `C.uint(level)` conversion. This will give you `^C.uint(0)`. If we pretend for a moment that `uint` and `C.uint` are the same, then we can see that [here](https://play.golang.org/p/TTmG7d7AvUy).

If `0xffffffff` is meaningless, then so is 9.

yuanqj8191: "Meaningless" I said is not proper here. It should be "clear". If I set level as 12, error will be "unsupported level 12". But if I set level as -3, it will be "unsupported level 4294967293". It not clear for the user.

kortschak: There is no error indicating the actual invalid value if it is discovered by HDF5. All we get back is an error code, -1, on failure, according to the HDF5 docs.

Please add the check for `> BestCompression`. When you have done that and sent a PR to gonum/gonum adding yourself to A+C with the commit message "A+C: add <yourfullname>" I will approve this PR. I don't want to waste more time here.

kortschak: Also the unsafe code above.

h5p_proplist.go
kortschak: Actually, we have this same pattern in other places.

@sbinet What do you want to do about this? There are many places in the code where there is type punning from `[]uint` to `[]C.hsize_t`. We can't guarantee that this type of unsafe conversion is correct.

I think that if we want to do it, there should be an initial `init` that checks this behaves as expected and panics informatively if not.

yuanqj8191: I already made a change. Please take a look.

sbinet: @kortschak _wrt_ the type prunning: you're probably right.
I've filed: https://github.com/gonum/hdf5/issues/49

sbinet:
@kortschak should we squash+merge or just rebase+merge?

kortschak:
I think this can be squashed to "hdf5: add zlib dataset compression support".

PR #54, opened 2019-09-12
Pull request to support dataset resize, HDF5 true colour image creation

zyc-sudo:
Hi,
I added three functionalities to the hdf5:
1. Added an utility of check if a node exist or not in the file
Problem: Other methods, for example, "open a group and check the err to determine if a group exist" works great until the call are done in a goroutine. For me it is a gRPC server method. Even if I checked the error. The C will still complaining the error( HDF5-DIAG errors).These error message won't have any affect on the goroutine's functionality since we handled the but they contaminated my log. Check this post 10 years ago: http://hdf-forum.184993.n3.nabble.com/check-if-dataset-exists-td194725.html.
Solution: looks like the best way is to use C's [H5Lexists](https://support.hdfgroup.org/HDF5/doc/RM/RM_H5L.html#Link-Exists). It is safe, no HDF5-DIAG errors anymore, and  works for group and dataset. 
So I added the following function in the h5g_group.go
`func (g *CommonFG) CheckLink(name string) error{}`

2. New support to create HDF5 true colour images
Problem: there is no wrapper for adding a true colour images for a CommonFG.( true colour images can be viewed in HDF5 compass as image directly instead of array.
Solution: Added the following function in the h5g_group.go
`func (g *CommonFG) CreateTrueImage(name string, img image.Image) error {}`

3. Added wrapper to resize a chunked dataset:
Problem: there is no wrapper for resize a chunked dataset
Solution: created the following function in h5d_dataset.go:
`func (s *Dataset) Resize(dims []uint) error {}`

P.S. I changed a few imports from `gonum.org/v1/hdf5` to `github.com/zyc-sudo/hdf5` just to make it work in my environment. Please don't merge those.

Please let me know if there are any improvements I need to add.

Regards,
Yucheng

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ matrix:
  allow_failures:
    - go: master
 
-go_import_path: gonum.org/v1/hdf5
+go_import_path: github.com/zyc-sudo/hdf5
 
 sudo: false
 
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # hdf5
+Fork from github.com/gonum/hdf5
 
 [![Build Status](https://secure.travis-ci.org/gonum/hdf5.png)](http://travis-ci.org/gonum/hdf5)
 [![GoDoc](https://godoc.org/gonum.org/v1/hdf5?status.svg)](https://godoc.org/gonum.org/v1/hdf5)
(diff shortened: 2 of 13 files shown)


zyc-sudo:
I

PR #57, opened 2019-09-17
Add a wrapper for C.H5Lexists for the CommonFG, returns nil if link e…

zyc-sudo:
Added the C.H5Lexists to check if the link(group, dataset, link) exists
Please take a look.

### DIFF
--- a/h5g_group.go
+++ b/h5g_group.go
@@ -11,7 +11,6 @@ package hdf5
 import "C"
 
 import (
-	"errors"
 	"fmt"
 	"unsafe"
 )
@@ -164,14 +163,9 @@ func (g *Group) OpenTable(name string) (*Table, error) {
 	return openTable(g.id, name)
 }
 
-// CheckLink checks if a link( can be group or dataset or actual link )exsit or not. This method won't give you annoying hdf5 warnings when called in a goroutine
-// CheckLink returns nil when a link exist, return an error when the link doesn't exist or some other error occured
-func (g *CommonFG) CheckLink(name string) error {
+// LinkExists returns whether a link with the specified name exists in the group.
+func (g *CommonFG) LinkExists(name string) bool {
 	c_name := C.CString(name)
 	defer C.free(unsafe.Pointer(c_name))
-	status := C.H5Lexists(g.id, c_name, 0)
-	if status > 0 {
-		return nil
-	}
-	return errors.New("The link " + name + " does not exist or some other error occured")
+	return C.H5Lexists(g.id, c_name, 0) > 0
 }
--- a/h5g_group.go
+++ b/h5g_group.go
@@ -11,6 +11,7 @@ package hdf5
 import "C"
 
 import (
+	"errors"
 	"fmt"
 	"unsafe"
 )
@@ -161,4 +162,16 @@ func (g *Group) CreateTableFrom(name string, dtype interface{}, chunkSize, compr
 // closed by the user when it is no longer needed.
 func (g *Group) OpenTable(name string) (*Table, error) {
 	return openTable(g.id, name)
+}
+
+// CheckLink checks if a link( can be group or dataset or actual link )exsit or not. This method won't give you annoying hdf5 warnings when called in a goroutine
+// CheckLink returns nil when a link exist, return an error when the link doesn't exist or some other error occured
+func (g *CommonFG) CheckLink(name string) error {
+	c_name := C.CString(name)
+	defer C.free(unsafe.Pointer(c_name))
+	status := C.H5Lexists(g.id, c_name, 0)
+	if status > 0 {
+		return nil
+	}
+	return errors.New("The link " + name + " does not exist or some other error occured")
 }
--- a/h5g_group_test.go
+++ b/h5g_group_test.go
@@ -17,8 +17,7 @@ func TestGroup(t *testing.T) {
 	defer os.Remove(fname)
 	defer f.Close()
 
-	err = f.CheckLink("/foo")
-	if err == nil {
+	if f.LinkExists("/foo") {
 		t.Fatalf("err: %s", "/foo shouldn't exist at this time")
 	}
 
@@ -33,14 +32,12 @@ func TestGroup(t *testing.T) {
 		t.Errorf("wrong Name for group: want %q, got %q", "/foo", g1.Name())
 	}
 
-	err = f.CheckLink("/foo")
-	if err != nil {
+	if !f.LinkExists("/foo") {
 		t.Fatalf("err: %s", err)
 	}
 
-	err = g1.CheckLink("bar")
-	if err == nil {
-		t.Fatalf("err: %s", "/foo shouldn't have bar as a child at this time")
+	if g1.LinkExists("bar") {
+		t.Fatalf("err: %s", "/foo shouldn't have bar as a child bar at this time")
 	}
 
 	g2, err := g1.CreateGroup("bar")
@@ -53,8 +50,7 @@ func TestGroup(t *testing.T) {
 	if g2.Name() != "/foo/bar" {
 		t.Errorf("wrong Name for group: want %q, got %q", "/foo/bar", g1.Name())
 	}
-	err = g1.CheckLink("bar")
-	if err != nil {
+	if !g1.LinkExists("bar") {
 		t.Fatalf("err: %s", err)
 	}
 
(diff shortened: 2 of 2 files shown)


zyc-sudo:
This GolangCI failure:    
compilation terminated.
    # gonum.org/v1/hdf5
    ./cgoflags.go:12:19: fatal error: hdf5.h: No such file or directory

I assume it is a CI setup error rather than code error?
     // #include "hdf5.h"

kortschak:
That failure looks to be related to https://github.com/golangci/golangci/issues/35. I'm not sure how to deal with this since we are completely tied to Cgo with this repo.

kortschak:
This should be fixed by #60.

zyc-sudo:
Great! Looking forward to contributing!

h5g_group.go
```diff
@@ -162,3 +163,15 @@ func (g *Group) CreateTableFrom(name string, dtype interface{}, chunkSize, compr
 func (g *Group) OpenTable(name string) (*Table, error) {
 	return openTable(g.id, name)
 }
+
+// CheckLink checks if a link( can be group or dataset or actual link )exsit or not. This method won't give you annoying hdf5 warnings when called in a goroutine
+// CheckLink returns nil when a link exist, return an error when the link doesn't exist or some other error occured
```
kortschak:
```
// LinkExists returns whether a link with the specified name exists in the group.
func (g *CommonFG) LinkExists(name string) bool {
	c_name := C.CString(name)
	defer C.free(unsafe.Pointer(c_name))
	return C.H5Lexists(g.id, c_name, 0) > 0
}
```

The HDF5 API is discusting for this. The return status, according to [this](https://support.hdfgroup.org/HDF5/doc/RM/RM_H5L.html#Link-Exists), is almost entirely uninformative for the failing case, so I'm not sure that it's even worth bothering to indicate the error return. @sbinet?

zyc-sudo: The reason why I chose an error instead of  bool is that in the h5group said" "Returns a negative value when the function fails and may return a negative value if the link does not exist". So looks like there are 3 cases: exist, doesn't exist, and some error. A bool is totally OK, exist or other than exist.

kortschak: >So looks like there are 3 cases: exist, doesn't exist, and some error. A bool is totally OK, exist or other than exist.

Yes, but in the linked docs, we see this 'Returns a negative value when the function fails and *may return a negative value if the link does not exist*. See “Failure Modes” below.', so negative may mean zero. Given that the API docs don't specify what the different negative values actually mean, it's not obvious how that information would be used (and in the current state of this PR, you already throw away the difference between these).

zyc-sudo: OK. That make sense. Do you like me to make the change to bool and push to this PR?

kortschak: Yes, please.

zyc-sudo:
Sure will do tomorrow

h5g_group.go
sbinet: yes, a `bool` might better reflect what's happening behind the curtains of the underlying C library.

zyc-sudo:
@kortschak Hi I make the changes, I also merged it to the other PR regarding the images where this feature is used

h5g_group_test.go
```diff
@@ -53,8 +50,7 @@ func TestGroup(t *testing.T) {
 	if g2.Name() != "/foo/bar" {
 		t.Errorf("wrong Name for group: want %q, got %q", "/foo/bar", g1.Name())
 	}
-	err = g1.CheckLink("bar")
-	if err != nil {
+	if !g1.LinkExists("bar") {
 		t.Fatalf("err: %s", err)
```
kortschak: There should not be and error here.

h5g_group_test.go
```diff
@@ -33,14 +32,12 @@ func TestGroup(t *testing.T) {
...
-	if err != nil {
+	if !f.LinkExists("/foo") {
 		t.Fatalf("err: %s", err)
 	}
 
-	err = g1.CheckLink("bar")
-	if err == nil {
-		t.Fatalf("err: %s", "/foo shouldn't have bar as a child at this time")
+	if g1.LinkExists("bar") {
+		t.Fatalf("err: %s", "/foo shouldn't have bar as a child bar at this time")
```
kortschak: This is not an error.

h5g_group_test.go
```diff
@@ -33,14 +32,12 @@ func TestGroup(t *testing.T) {
 		t.Errorf("wrong Name for group: want %q, got %q", "/foo", g1.Name())
 	}
 
-	err = f.CheckLink("/foo")
-	if err != nil {
+	if !f.LinkExists("/foo") {
 		t.Fatalf("err: %s", err)
```
kortschak: There is no error here.

h5g_group_test.go
```diff
@@ -17,8 +17,7 @@ func TestGroup(t *testing.T) {
 	defer os.Remove(fname)
 	defer f.Close()
 
-	err = f.CheckLink("/foo")
-	if err == nil {
+	if f.LinkExists("/foo") {
 		t.Fatalf("err: %s", "/foo shouldn't exist at this time")
```
kortschak: There is no error here.

zyc-sudo: This one should be an error because if the /foo already existed before it was created than that is not quite right

h5g_group_test.go
zyc-sudo: OOPS, sorry about this. Fixed it in the new commit

h5g_group_test.go
kortschak: The condition being tested is not an error condition.
```
t.Fatal("unexpected /foo link present")
```

h5g_group_test.go
```diff
@@ -28,6 +32,14 @@ func TestGroup(t *testing.T) {
 		t.Errorf("wrong Name for group: want %q, got %q", "/foo", g1.Name())
 	}
 
+	if !f.LinkExists("/foo") {
+		t.Fatalf("err: %s", "/foo should exist at this time")
```
kortschak: Please remove the "err:" prefixes on these test failures. They don't add anything and they are misleading.

```
t.Error(`unexpected "/foo" group link`)
```

h5g_group_test.go
kortschak:
```
t.Error(`unexpected "bar" child link for "/foo" group`)
```

h5g_group_test.go
```diff
@@ -28,6 +32,14 @@ func TestGroup(t *testing.T) {
 		t.Errorf("wrong Name for group: want %q, got %q", "/foo", g1.Name())
```
kortschak:
```
t.Errorf(`wrong name for group: want:"/foo", got:%q`, g1.Name())
```

h5g_group_test.go
```diff
@@ -17,6 +17,10 @@ func TestGroup(t *testing.T) {
 	defer os.Remove(fname)
 	defer f.Close()
 
+	if f.LinkExists("/foo") {
+		t.Fatal("unexpected /foo link present")
```
kortschak:
```
t.Error("unexpected /foo link present")
```

h5g_group_test.go
```diff
@@ -38,6 +50,9 @@ func TestGroup(t *testing.T) {
 	if g2.Name() != "/foo/bar" {
 		t.Errorf("wrong Name for group: want %q, got %q", "/foo/bar", g1.Name())
 	}
+	if !g1.LinkExists("bar") {
+		t.Fatalf("err: %s", "/foo shouldn't have bar as a child bar at this time")
```
kortschak:
```
t.Error(`expected "bar" child link for "/foo" group not present`)
```

h5g_group_test.go
zyc-sudo: Shouldn't this line 36 be 

> t.Error(`expected "/foo" group link`)

kortschak: Yes. Thanks.

zyc-sudo:
Just pushed all the requested changes

kortschak:
Thanks

zyc-sudo:
Thanks a lot! This is the first time I participated github open source.
Thank you very much for your patience

kortschak:
I neglected to say, would you please send a PR adding yourself to AUTHORS and CONTRIBUTORS at gonum/gonum. Please use the commit message "A+C: add Yucheng Zhu" (I assume this is your name from the commits).

PR #64, opened 2019-10-04
This change modifies Gonum to support Slice in Struct

egachen:
This change modifies Gonum to support Slice in Struct

### DIFF
--- a/cmem/encoder.go
+++ b/cmem/encoder.go
@@ -72,6 +72,7 @@ func (enc *Encoder) Encode(data interface{}) error {
 	}
 
 	rt := rv.Type()
+	fmt.Println("Kind: ", rt.Kind())
 
 	switch rt.Kind() {
 	case reflect.Slice:
--- a/cmem/encoder.go
+++ b/cmem/encoder.go
@@ -4,6 +4,11 @@
 
 package cmem
 
+// #include "hdf5.h"
+// #include <stdlib.h>
+// #include <string.h>
+import "C"
+
 import (
 	"encoding/binary"
 	"fmt"
@@ -26,8 +31,16 @@ func init() {
 // subsequently write an in memory object to a PacketTable using Append().
 type Encoder struct {
 	// Buf contains the encoded data.
-	Buf    []byte
-	offset int
+	Buf          []byte
+	offset       int
+	pointerSlice []unsafe.Pointer
+}
+
+// FreeMemory is to free memory from C
+func (enc *Encoder) FreeMemory() {
+	for i := 0; i < len(enc.pointerSlice); i++ {
+		C.free(enc.pointerSlice[i])
+	}
 }
 
 // Encode encodes the value passed as data to binary form stored in []Buf. This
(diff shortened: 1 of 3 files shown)


cmem/encoder_test.go
```diff
@@ -6,51 +6,50 @@ package cmem
...
 	}
+	if !bytes.Equal(enc.Buf, want) {
+		t.Fatalf("encoding error:\ngot = %v\nwant= %v", enc.Buf, want)
+	}
+}
+
+func TestEncodeSlice(t *testing.T) {
+	v := []int32{1, 20, 300, 4000, 50000, 600000, 7000000, 8000000}
+	want := []byte{
+		0x08, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, // |........|
```
kortschak: This was manually created and is not correct.

cmem/encoder_test.go
```diff
@@ -6,51 +6,50 @@ package cmem
...
 import (
 	"bytes"
-	"encoding/binary"
-	"strconv"
 	"testing"
+	"encoding/binary"
 )
 
 func TestEncode(t *testing.T) {
-	for i, tc := range []struct {
```
kortschak: Why was this deleted?

h5pt_table.go
```diff
@@ -167,6 +169,7 @@ func createTable(id C.hid_t, name string, dtype *Datatype, chunkSize, compressio
 	if err := checkID(hid); err != nil {
 		return nil, err
 	}
+
```
kortschak: Why?

cmem/encoder_test.go
egachen: Based on the HDF5 hvl_t strcture, 
/* Variable Length Datatype struct in memory */
/* (This is only used for VL sequences, not VL strings, which are stored in char *'s) */
typedef struct {
    size_t len; /* Length of VL data (in base type units) */
    void *p;    /* Pointer to VL data */
} hvl_t;

The variable length array in memory has two fields: len and pointer to the heap memory which is holding the actual data.

In this test case, the first 8 bytes indicates the length of the data: 8

cmem/encoder_test.go
egachen: The original test case was not correct for variable length data, it's supposed to handle it same as fix length array. So I removed the second test data, and the loop is not needed anymore

cmem/encoder.go
```diff
@@ -61,7 +73,39 @@ func (enc *Encoder) Encode(data interface{}) error {
 	rt := rv.Type()
 
 	switch rt.Kind() {
-	case reflect.Slice, reflect.Array:
+	case reflect.Slice:
+		length := C.size_t(reflect.ValueOf(data).Len())
+		msize := C.size_t(rv.Index(0).Type().Size() * uintptr(length))
+
+		// this is variable length data, it should follow hvl_t format
+		enc.Encode(length)
```
golangcibot: Error return value of `enc.Encode` is not checked (from `errcheck`)

cmem/encoder.go
```diff
@@ -61,7 +73,39 @@ func (enc *Encoder) Encode(data interface{}) error {
...
+			if mypad > 0 {
+				myenc.Buf = append(myenc.Buf, make([]byte, mypad)...)
+			}
+
+			tempBuf = append(tempBuf, myenc.Buf...)
+		}
+		// copy contents from temp buf to C memory
+		C.memcpy(pointer, unsafe.Pointer(&tempBuf[0]), msize)
+
+		enc.Encode(C.size_t(uintptr(pointer)))
```
golangcibot: Error return value of `enc.Encode` is not checked (from `errcheck`)

PR #66, opened 2019-11-04
ci: add Go-1.13, drop Go-1.10

sbinet:
Please take a look.


sbinet:
Sure.
I'll do it when in front of a real keyboard.

PR #70, opened 2020-05-01
Adds support for aarch64

noj-richards:
This change modifies hdf5 so that it compiles and links on aarch64 architectures.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,10 @@ go:
   - 1.12.x
   - master
 
+arch:
+  - amd64
+  - arm64
+
 matrix:
  fast_finish: true
  allow_failures:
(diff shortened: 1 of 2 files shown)


kortschak:
I have confirmed that this does what it says, be we need to have this tested on Travis as well. Can you please add an arch line to `.travis.yml` so that this happens (like [here](https://github.com/gonum/gonum/blob/015024673d2bbd05971e5baac93843f2c180e792/.travis.yml#L17-L19)).

kortschak:
Also, please send a PR to gonum/gonum adding yourself (/employer if relevant) to AUTHORS and CONTRIBUTORS with the commit message "A+C: add Jon Richards". We will merge that when this is merged.

noj-richards:
Thanks. Both requests now complete.

PR #78, opened 2021-03-08
hdf5: add ObjectTypeByIndex to CommonFG

### LINKED ISSUE #77: Is there a way to know the type of the object without trying to open first?
### What are you trying to do?
I'm reading this file: https://github.com/rcmalli/keras-vggface/releases/download/v2.0/rcmalli_vggface_tf_notop_resnet50.h5

### What actually happened?
I'm using a code based on the code listed in #76 to read the file listed above, the problem is that this file has group inside group but I always try to read datasets inside groups so it fails.
Is there a way to know that the object is a group or dataset?

What I did to solve this was attempting to open a group first if it fails it attempts to open a dataset but I wonder if there's a cleaner way to do this

### What version of Go, Gonum, Gonum/netlib and libhdf5 are you using?

gonum.org/v1/hdf5: master
go version: 1.16

### Does this issue reproduce with the current master?
yes

sbinet:
This CL adds a new method to CommonFG, ObjectTypeByIndex, to retrieve the type of an object in
a file or a group, prior to read from storage.

Fixes #77.

### DIFF
--- a/h5g_group.go
+++ b/h5g_group.go
@@ -15,6 +15,37 @@ import (
 	"unsafe"
 )
 
+// GType describes the type of an object inside a Group or File.
+type GType int
+
+const (
+	H5G_UNKNOWN GType = C.H5G_UNKNOWN // Unknown object type
+	H5G_GROUP   GType = C.H5G_GROUP   // Object is a group
+	H5G_DATASET GType = C.H5G_DATASET // Object is a dataset
+	H5G_TYPE    GType = C.H5G_TYPE    // Object is a named data type
+	H5G_LINK    GType = C.H5G_LINK    // Object is a symbolic link
+	H5G_UDLINK  GType = C.H5G_UDLINK  // Object is a user-defined link
+)
+
+func (typ GType) String() string {
+	switch typ {
+	case H5G_UNKNOWN:
+		return "unknown"
+	case H5G_GROUP:
+		return "group"
+	case H5G_DATASET:
+		return "dataset"
+	case H5G_TYPE:
+		return "type"
+	case H5G_LINK:
+		return "link"
+	case H5G_UDLINK:
+		return "udlink"
+	default:
+		return fmt.Sprintf("GType(%d)", int(typ))
+	}
+}
+
 // CommonFG is for methods common to both File and Group
 type CommonFG struct {
 	Identifier
@@ -143,6 +174,17 @@ func (g *CommonFG) ObjectNameByIndex(idx uint) (string, error) {
 		return "", fmt.Errorf("could not get name")
 	}
 	return C.GoString(&name[0]), nil
+}
+
+// ObjectTypeByIndex returns the type of the object at idx.
+func (g *CommonFG) ObjectTypeByIndex(idx uint) (GType, error) {
+	cidx := C.hsize_t(idx)
+	gtyp := GType(C.H5Gget_objtype_by_idx(g.id, cidx))
+	if gtyp < H5G_GROUP {
+		return H5G_UNKNOWN, fmt.Errorf("could not get object type")
+	}
+
+	return gtyp, nil
 }
 
 // CreateTable creates a packet table to store fixed-length packets.
--- a/h5i_identifier.go
+++ b/h5i_identifier.go
@@ -11,6 +11,7 @@ package hdf5
 import "C"
 
 import (
+	"fmt"
 	"unsafe"
 )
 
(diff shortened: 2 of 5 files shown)


sbinet:
not completely happy with the naming of the new GType constants...
happy to get some input there.

h5g_group.go
```diff
@@ -145,6 +173,17 @@ func (g *CommonFG) ObjectNameByIndex(idx uint) (string, error) {
 	return C.GoString(&name[0]), nil
 }
 
+// ObjectTypeByIndex returns the type of the object at idx.
+func (g *CommonFG) ObjectTypeByIndex(idx uint) (GType, error) {
+	cidx := C.hsize_t(idx)
+	ctyp := C.H5Gget_objtype_by_idx(g.id, cidx)
+	if ctyp < 0 {
```
kortschak, line 180: Does `GType(0)` have a name?

h5i_identifier.go
```diff
@@ -26,6 +27,27 @@ const (
...
+	case DATASPACE:
+		return "dataspace"
+	case DATASET:
+		return "dataset"
+	case ATTRIBUTE:
+		return "attribute"
+	case BAD_ID:
+		return "bad_id"
+	default:
+		return fmt.Sprintf("type=%d", int(typ))
```
kortschak, line 47: `IType(%d)` for consistency with `GType`?

h5g_group.go
sbinet: yes, that's `H5G_GROUP`.
I'll update that.

dcu:
what's wrong with the naming?

sbinet:
I wasn't completely happy with the `H5G_GROUP` naming as, from the user point of view, the stuttering isn't pretty (`hdf5.H5G_GROUP`...).

a better naming scheme would be nice.

anyways... PTAL?

kortschak:
We can just blame it on the HDF Group; I think it's reasonable to keep constants in line with the names they have there (ugly as they may be) to ease finding of documentation.

h5g_group.go
```diff
@@ -15,6 +15,34 @@ import (
...
+// GType describes the type of an object inside a Group or File.
+type GType byte
+
+const (
+	H5G_GROUP   = GType(C.H5G_GROUP)
+	H5G_DATASET = GType(C.H5G_DATASET)
+	H5G_TYPE    = GType(C.H5G_TYPE)
+	H5G_LINK    = GType(C.H5G_LINK)
+	H5G_UDLINK  = GType(C.H5G_UDLINK)
+)
```
kortschak, line 27: Use the same style as for the `IType` constants. I think that with this syntax the consts don't end up associated with the type in the documentation.

h5g_group.go
```diff
@@ -145,6 +173,17 @@ func (g *CommonFG) ObjectNameByIndex(idx uint) (string, error) {
 	return C.GoString(&name[0]), nil
 }
 
+// ObjectTypeByIndex returns the type of the object at idx.
+func (g *CommonFG) ObjectTypeByIndex(idx uint) (GType, error) {
+	cidx := C.hsize_t(idx)
+	gtyp := GType(C.H5Gget_objtype_by_idx(g.id, cidx))
```
kortschak, line 179: Apparently this is deprecated in 1.8 in favour of H5Oget_info (see https://support.hdfgroup.org/HDF5/doc/RM/RM_H5G.html#Group-GetObjTypeByIdx).

sbinet: indeed.
I was aiming at the most common denominator.

I am not sure we have an explicit policy here:
- should we follow deprecated notices and thus tighly follow the latest 1.X releases of C-HDF5?
- or rather be the most conservative?
- or give up and provide a way to select which API to wrap (adding more work on our side to maintain multiple APIs and dropping the `go-get`-able aspect of that library. granted that with `Cgo` in the mix, `go-get`-ablity is always a somewhat fluid notion...)

alternatively, we could also vendor the C-HDF5 library we wrap.

h5g_group.go
kortschak: Yep

h5g_group.go
dcu: vendoring the hdf5 would be better I think

dcu (approved):
tested this with my code and worked well

go 1.16
hdf5 1.12.0

h5g_group.go
sbinet: @kortschak do you have recommandations about a policy?
(as noted in the thread on `gonum-dev`, vendoring is a rather big chunk of work - for which I wouldn't have the cycles to spare at the moment)

kortschak: If someone else were to do the work, we could review, but I'm not particularly inclined to do it myself. That's not a policy, but it's based on how much time there in available from maintainers at the moment.

sbinet: I'd help with the review as well.

but, concerning this PR, what should we do?

h5g_group.go
```diff
@@ -145,6 +173,17 @@ func (g *CommonFG) ObjectNameByIndex(idx uint) (string, error) {
...
+	if gtyp < H5G_GROUP {
+		return 0, fmt.Errorf("could not get object type")
```
kortschak, line 181: Is there a const for this 0?

h5g_group.go
kortschak: I think we can merge this and kick the can down the road.

h5g_group.go
sbinet: yes, opted for `H5G_UNKNOWN`

sbinet:
PTAL

kortschak (approved):

